PR #1, opened 2021-01-28
Redis operator

edumucelli:
This PR adds the initial version of the Redis Operator for K8s. It was based on the [mongodb](https://github.com/canonical/mongodb-operator), [elasticsearch](https://github.com/canonical/elasticsearch-operator/) and [this PR](https://github.com/canonical/charm-mssql/pull/6/files).

It was made with the goal to first serve for the Commercial Systems team to deploy a Redis cache without replication. Therefore, at first it does not provide Redis cluster capabilities.

As a newbie on the operator framework, juju and at certain point k8s I still uncertain of parts of this code. I have left a comment on each of them. If anyone knows, I'd appreciate any further clarification.


src/pod_spec.py
```diff
@@ -0,0 +1,110 @@
...
+    def _build_env_conf_spec(self) -> Dict:
+        config_fields = {
+            "JUJU_NODE_NAME": "spec.nodeName",
+            "JUJU_POD_NAME": "metadata.name",
+            "JUJU_POD_NAMESPACE": "metadata.namespace",
+            "JUJU_POD_IP": "status.podIP",
+        }
+        env_config = {k: {"field": {"path": p, "api-version": "v1"}} for k, p in config_fields.items()}
+
+        env_config["JUJU_EXPECTED_UNITS"] = " ".join(self.expected_units())
```
edumucelli: I could not find documentation around those Juju env variables. In doubt, I have added them. I have found them in [Postgresql](https://jaas.ai/u/postgresql-charmers/postgresql-k8s) charm. What are they for?

src/pod_spec.py
```diff
@@ -0,0 +1,110 @@
...
+        if not image_info:
+            image_info = {}
+        self.name = name
+        self.port = port
+        self.image_info = image_info
+
+    def build_pod_spec(self) -> Dict:
+        """Set up and return our full pod spec."""
+
+        # vol_config = [
```
edumucelli, line 41: I am not sure if a volume configuration is necessary?

README.md
```diff
@@ -0,0 +1,92 @@
...
+Then you can 
+
+    redis-cli -h 10.1.168.69 -p 3000
+
+Another option is to port-forward the Redis service (or pod) port from k8s to a local port.
+For example, forwarding a node's port
+
+    microk8s.kubectl --namespace=redis-model port-forward pod/redis-8485b69dfd-6hgvj 6379:6379
+
+However, it is simpler to forward the service port as it will abstract the node name behind it
```
alnvdl: `:` missing at the end of this line and others below.

README.md
```diff
@@ -0,0 +1,92 @@
...
+Create and activate a virtualenv with the development requirements:
+
+    virtualenv -p python3 venv
+    source venv/bin/activate
+    pip install -r requirements-dev.txt
+
+In order to check the result of a modification, rebuild and upgrade the charm
+
+    charmcraft build
+    juju upgrade-charm --path="/home/eduardo/Canonical/redis-k8s/redis.charm" redis --force-units
```
alnvdl: Use a relative path instead, or abstract it?

metadata.yaml
```diff
@@ -0,0 +1,55 @@
...
+summary: >
+  Redis is an open source (BSD licensed), in-memory data structure store, used
+  as a database, cache, and message broker. Redis provides data structures
+  such as strings, hashes, lists, sets, sorted sets with range queries,
+  bitmaps, hyperloglogs, geospatial indexes, and streams. Redis has built-in
+  replication, Lua scripting, LRU eviction, transactions, and different levels
+  of on-disk persistence, and provides high availability via Redis Sentinel and
+  automatic partitioning with Redis Cluster.
+
+  This charm supports high Redis 6 in Kubernetes environments, using k8s services
```
alnvdl: What is high Redis?

run_tests
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -e
+# Copyright 2021 Eduardo Mucelli Rezende Oliveira
```
alnvdl: Since it's under Canonical's org, perhaps use Canonical here as in the other files?

src/client.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class RedisClient:
+    def __init__(self, host: str, port: int):
+        self.host = host
+        self.port = port
+
+    def is_ready(self) -> bool:
+        try:
+            redis.Redis(host=self.host, port=self.port)
```
alnvdl: Shouldn't you close this after you're done with it?

src/pod_spec.py
alnvdl: I think it would be if you stored the Redis persistence file to some volume so that it persists beyond the pod lifecycle? I think it could probably come as an improvement to this charm.

alnvdl (approved):
LGTM!

src/pod_spec.py
cmars: This seems like some Juju internals poking through the operator framework. Seems strange to manipulate JUJU_ agent environment variables directly.

I'd suggest going the other way.. leaving it out and adding it when you find what breaks :)

That way we know what it's for and can comment. Or maybe we never needed it; PostgreSQL clustering is probably quite different from the way Redis does it.

It's also possible that the PostgreSQL k8s charm was developed against a pre-release of the operator charm (which would also explain that internals-poking-through vibe).

metadata.yaml
```diff
@@ -0,0 +1,55 @@
...
+  service: loadbalancer
+storage:
+  database:
+    type: filesystem
+    location: /var/opt/redis
+resources:
+  redis-image:
+    type: oci-image
+    description: upstream docker image for redis
+    upstream-source: 'redis:6.0'
```
cmars: Missing newline

metadata.yaml
```diff
@@ -0,0 +1,55 @@
...
+  This charm supports high Redis 6 in Kubernetes environments, using k8s services
+  for load balancing.
+maintainers:
+  - Redis Charm Maintainers <redis-charmers@lists.launchpad.net>
+min-juju-version: 2.8.7
+series:
+  - kubernetes
+provides:
+  db:
+    interface: redis
```
cmars: It might not be a good idea to declare a relation endpoint before it is implemented. Unless there's a followup to add that. Maybe a TODO comment?

README.md
```diff
@@ -0,0 +1,92 @@
...
+To access Redis you run
+
+    juju status
+
+to discover the IP Redis is running behind. The output will have lines like
+
+    Unit       Workload    Agent  Address       Ports     Message
+    redis/20   active      idle   10.1.168.69   6379/TCP  Pod is ready.
+
+Then you can
```
cmars: It might be useful to document how to connect to the service if you're trying to connect from other services that aren't charmed up yet. Like what kube dns name & port to expect. Hopefully the operator framework can interoperate with existing k8s...

src/client.py
cmars: Not familiar with the Python client. Does this actually do anything like connect via TCP? Should you do a [PING](https://redis.io/commands/ping) here?

src/pod_spec.py
cmars: +1, I'd leave a TODO for persistence. Not all redis use cases need persistence but some will expect it so it's something to mention in the README.

src/pod_spec.py
```diff
@@ -0,0 +1,110 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Dict, List
+
+from charmhelpers.core import hookenv
+
+
+class PodSpecBuilder:
```
cmars, line 19: Docstring?

src/charm.py
```diff
@@ -0,0 +1,157 @@
...
+    def update_status(self, _):
+        """Set status for all units
+
+        Status may be
+        - Redis API server not reachable (service is not ready),
+        - Ready
+        """
+        if not self.unit.is_leader():
+            self.unit.status = ActiveStatus('Pod is ready.')
+            return
```
cmars, line 148: If you return early here, it looks like that means non-leaders would not check is_ready below...

Not sure how per-unit vs per-app statuses are surfaced so maybe it's fine? Something to consider anyway.

metadata.yaml
```diff
@@ -0,0 +1,55 @@
+# This file is part of the Redis k8s Charm for Juju.
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3, as
+# published by the Free Software Foundation.
```
cmars, line 6: Probably need a LICENSE or copyright file (whatever is the convention for operator charms) with GPLv3.

metadata.yaml
cmars: Pinning a version tag is wise.

cmars (approved):
Looks good and I enjoyed reading this, as a crash course in operator charms. A few thoughts below.

Also given that this project is in the public canonical org, the README should be up front about what is and isn't supported for future travelers here.

README.md
edumucelli: Thanks! Done!

README.md
edumucelli: Yes, good idea, I kind of let it there and even forgot, because I took it directly from my notes while I was learning Juju. Added a note considering that you in the directory of the project.

run_tests
edumucelli: Ouch, forgot to change this! Good catch!

metadata.yaml
edumucelli: "kkkkkkk", it is something I just made up.

src/client.py
edumucelli: Good catch, have done it and associated it with the charm `on_stop` event. Thanks!

metadata.yaml
edumucelli: Thanks, yes, I was not sure also here which way to go. I have removed it. Honestly, I do not fully understand yet what are the consequences of providing that and how to manage its lifecycle in the operator.

metadata.yaml
edumucelli: Thanks, good catch, added it.

README.md
edumucelli: Good idea, I have hopefully clarified that. Thanks!

src/pod_spec.py
edumucelli: Thanks, added!

metadata.yaml
edumucelli: :+1: , I have seen others doing the same and just stating in their doc that they support X version of a service. Did the same.

src/charm.py
edumucelli: Yes, I will look into that. I kind of went with the other implementations I have seen. The first thing they have checked was that, even those that required the non-leaders to be ready on the rest of the flow. I will investigate that by adding units. It is a very good point! Thanks!

src/pod_spec.py
edumucelli: Removed them.

src/pod_spec.py
edumucelli: Commented as `TODO`

edumucelli:
Thanks @alnvdl and @cmars for the review so far. One common thing you have commented in common was the `volume` thing I was not sure about. I will let a `TODO` there so that if we need it in the future we know directly where it should go.

@cmars I have[ removed the `JUJU` variables](https://github.com/canonical/redis-operator/pull/1#discussion_r566504748) then. It was also something @balbirthomas suggested me to do in a direct message.

@cmars I have added the `PING` to actually check the communication with Redis as your suggestion. By instantiating the client already establishes the connection, but the `PING` actually calls it, so that at least has the extra action of using the connection pool and being explicit in the code.

metadata.yaml
edumucelli: There is a LICENSE file already in the repository, I have started the repository with it, that is why it is not in the PR. As you have mentioned it is GPLv3.

src/charm.py
edumucelli: After trying some scenarios, yes, the behavior you described does happen. I am now wondering why [elasticsearch](https://github.com/canonical/elasticsearch-operator/blob/790a58e30b4aaa3621a152320e5d66c7169ed14b/src/charm.py#L362-L364) and [mongo](https://github.com/canonical/mongodb-operator/blob/fdc1fe208b7539e100c959f7ae6eaaa773d17733/src/charm.py#L159-L161) operators do it that way. Maybe because the connection through the service will be made by the leader (by some magic I do not know) so we do not care about the status of the other units and set active for simplicity?

edumucelli: @balbirthomas I'd really appreciate your view on this discussion, I lack more knowledge here to understand why it is done that way for non-leader units. In the case of this operator here there won't have support for clustering, so having more than 1 unit is not useful. Maybe there is a better way to handle it. Any suggestions are welcome.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    branches: [ develop, main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
```
frankban: Can we drop 3.5? Bionic has 3.6 and I guess we could just target focal.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  push:
+    branches: [ develop, main ]
+  pull_request:
+    branches: [ develop, main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
frankban: Should we run on lts?

.jujuignore
```diff
@@ -0,0 +1,6 @@
+/venv
+*.py[cod]
+*.charm
+build/
+__pycache__
+.idea
```
frankban: Missing newline at end of file.

README.md
```diff
@@ -0,0 +1,111 @@
...
+To see the logs for a specific pod:
+    
+    microk8s.kubectl --namespace=redis-model logs redis-operator-0
+
+## Testing
+
+The Python operator framework includes a very nice harness for testing
+operator behaviour without full deployment. Just `run_tests`:
+
+    ./run_tests
```
frankban, line 112: This is great docs! Thanks!

metadata.yaml
```diff
@@ -0,0 +1,52 @@
...
+  such as strings, hashes, lists, sets, sorted sets with range queries,
+  bitmaps, hyperloglogs, geospatial indexes, and streams. Redis has built-in
+  replication, Lua scripting, LRU eviction, transactions, and different levels
+  of on-disk persistence, and provides high availability via Redis Sentinel and
+  automatic partitioning with Redis Cluster.
+
+  This charm supports Redis 6 in Kubernetes environments, using k8s services
+  for load balancing. This charm does not support Redis clustering capabilities.
+maintainers:
+  - Redis Charm Maintainers <redis-charmers@lists.launchpad.net>
```
frankban: Who owns this email?

requirements.txt
```diff
@@ -0,0 +1,22 @@
...
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranties of
+# MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+ops~=1.1.0
```
frankban, line 18: Put those in alphabetical order?

requirements.txt
```diff
@@ -0,0 +1,22 @@
...
+redis~=3.5.3
+
```
frankban: Why the newline?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+# WITHOUT ANY WARRANTY; without even the implied warranties of
+# MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+
+import yaml
```
frankban, line 20: Alphabetical order.
To avoid those things we could use black and isort as done in the salesborg project. Can be a follow-up branch.

README.md
```diff
@@ -0,0 +1,111 @@
...
+    microk8s.kubectl --namespace=redis-model port-forward service/redis 6379:6379
+
+Then you can simply:
+
+    redis-cli
+
+From a k8s non-charmed scenario Redis will be exposed as `Service` named `redis` on the default
+port `6379`.
+
+## Developing
```
frankban, line 68: Could you please add a link to the operator framework in this paragraph?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        self.framework.observe(self.on.config_changed, self.configure_pod)
+        self.framework.observe(self.on.upgrade_charm, self.configure_pod)
+        self.framework.observe(self.on.update_status, self.update_status)
+
+    def on_start(self, event):
+        """Initialize Redis
+
+        This event handler is deferred if initialization of Redis fails.
+        """
+        logger.debug("Running on_start")
```
frankban: I wonder if a decorator like that could work:
```python
def log_event_handler(method):
    @functools.wraps(method)
    def decorated(self, event):
        logger.debug(f"running {method.__name__}")
        try:
            return method(self, event)
        finally:
            logger.debug(f"completed {method.__name__}")
    return decorated
```
And then you could just:
```python
@log_event_handler
def on_start(self, event):
```
And so on.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        self.image = OCIImageResource(self, "redis-image")
+
+        self.framework.observe(self.on.start, self.on_start)
+        self.framework.observe(self.on.stop, self.on_stop)
+        self.framework.observe(self.on.config_changed, self.configure_pod)
+        self.framework.observe(self.on.upgrade_charm, self.configure_pod)
+        self.framework.observe(self.on.update_status, self.update_status)
+
+    def on_start(self, event):
+        """Initialize Redis
```
frankban: Final dot.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+
+        This event handler is deferred if initialization of Redis fails.
+        """
+        logger.debug("Running on_start")
+        if not self.unit.is_leader():
+            self.unit.status = ActiveStatus()
```
frankban, line 76: I guess this is the line you had questions about. I agree this is not clear what it does.

src/client.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+
+import redis
+
+from log_adapter import CustomAdapter
+
```
frankban, line 21: Please add two newlines between imports and code.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+from oci_image import OCIImageResource, OCIImageResourceError
+from ops.charm import CharmBase
+from ops.framework import StoredState
+from ops.main import main
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+
+from client import RedisClient
+from log_adapter import CustomAdapter
+from pod_spec import PodSpecBuilder
+
```
frankban, line 30: Please add two newlines between imports and code.

src/client.py
```diff
@@ -0,0 +1,48 @@
...
+            if self.redis.ping():
+                logger.debug("We can ping Redis, service is ready.")
+                return True
+            logger.debug("Not able to ping Redis.")
+            return False
+        except redis.exceptions.ConnectionError as exc:
+            logger.warning("Unable to connect to Redis: {}".format(exc))
+        return False
+
+    def _get_client(self) -> redis.Redis:
```
frankban: Is this method really needed? It's only called once it seems.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            return
+
+        self.pod_is_ready()
+
+    def pod_is_ready(self):
+        logger.debug('Pod is ready.')
+        self.unit.status = ActiveStatus()
+        self.app.status = ActiveStatus()
+
+    @property
```
frankban: With redis being a property, we are instantiating it every time. Can it be a regular attribute?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            event.defer()
+            return
+
+        self.pod_is_ready()
+        logger.debug("Running on_start finished")
+
+    def on_stop(self, _):
+        """Mark terminating unit as inactive
+        """
+        self.redis.close()
```
frankban, line 91: Is this actually doing anything at all, given that redis is a property?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            self.unit.status = WaitingStatus(WAITING_FOR_REDIS_MSG)
+            logger.debug("{}: deferring on_start".format(WAITING_FOR_REDIS_MSG))
+            event.defer()
+            return
+
+        self.pod_is_ready()
+        logger.debug("Running on_start finished")
+
+    def on_stop(self, _):
+        """Mark terminating unit as inactive
```
frankban: Final dot.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            self.unit.status = ActiveStatus()
+            return
+
+        if not self.redis.is_ready():
+            self.unit.status = WaitingStatus(WAITING_FOR_REDIS_MSG)
+            logger.debug("{}: deferring on_start".format(WAITING_FOR_REDIS_MSG))
+            event.defer()
+            return
+
+        self.pod_is_ready()
```
frankban: This feels like a function returning a boolean. Maybe `set_ready_status`?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        logger.debug("Running on_start finished")
+
+    def on_stop(self, _):
+        """Mark terminating unit as inactive
+        """
+        self.redis.close()
+        self.unit.status = MaintenanceStatus('Pod is terminating.')
+
+    def configure_pod(self, event):
+        """Applies the pod configuration
```
frankban: Final dot.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        if not self.unit.is_leader():
+            logger.debug("Spec changes ignored by non-leader")
+            self.unit.status = ActiveStatus()
+            return
+
+        msg = 'Configuring pod.'
+        logger.debug(msg)
+        self.unit.status = WaitingStatus(msg)
+
+        # Fetch image information
```
frankban: Final dot.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        logger.debug("Running configure_pod")
+
+        if not self.unit.is_leader():
+            logger.debug("Spec changes ignored by non-leader")
+            self.unit.status = ActiveStatus()
+            return
+
+        msg = 'Configuring pod.'
+        logger.debug(msg)
+        self.unit.status = WaitingStatus(msg)
```
frankban: This status message would only be relevant until next line, is it worth it?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            image_info=image_info,
+        )
+
+        spec = builder.build_pod_spec()
+        logger.debug("Pod spec: \n{}".format(yaml.dump(spec)))
+
+        # Update pod spec if the generated one is different
+        # from the one previously applied
+        if self.state.pod_spec == spec:
+            logger.debug("Discarding pod spec because it has not changed.")
```
frankban, line 124: Nice!

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            name=self.model.app.name,
+            port=DEFAULT_PORT,
+            image_info=image_info,
+        )
+
+        spec = builder.build_pod_spec()
+        logger.debug("Pod spec: \n{}".format(yaml.dump(spec)))
+
+        # Update pod spec if the generated one is different
+        # from the one previously applied
```
frankban: Final dot.

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        if self.state.pod_spec == spec:
+            logger.debug("Discarding pod spec because it has not changed.")
+        else:
+            logger.debug("Applying new pod spec.")
+            self.model.pod.set_spec(spec)
+            self.state.pod_spec = spec
+
+        if not self.redis.is_ready():
+            self.unit.status = WaitingStatus(WAITING_FOR_REDIS_MSG)
```
frankban, line 131: This code here seems to be repeated multiple times, maybe it's worth abstracting this out as a method?

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+            logger.debug("{}: deferring configure_pod".format(WAITING_FOR_REDIS_MSG))
+            event.defer()
+            return
+
+        self.pod_is_ready()
+        logger.debug("Running configure_pod finished")
+
+    def update_status(self, _):
+        """Set status for all units
```
frankban: Final dot.

src/charm.py
frankban: Also, the redis client here seems to only be needed to check whether we can connect to redis. So I guess all we actually need is a `is_redis_ready` function that can be called by the charm, without exposing the underlying redis object.

src/client.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+
+import redis
+
+from log_adapter import CustomAdapter
+
+logger = CustomAdapter(logging.getLogger(__name__), {'prefix': 'redis-operator:client'})
+
+
+class RedisClient:
```
frankban, line 25: Please add a docstring. Also, as mentioned, consider whether a simple function checking we can connect to redis would be better than this client wrapper.

src/client.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, host: str, port: int):
+        self.host = host
+        self.port = port
+        self.redis: redis.Redis = None
+
+    def is_ready(self) -> bool:
```
frankban, line 31: Docstring.

src/log_adapter.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+
+
+class CustomAdapter(logging.LoggerAdapter):
+    """
+    This adapter prefix the log messages by a given 'prefix' key
```
frankban: Final dot.

src/log_adapter.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+
+    def process(self, msg, kwargs):
```
frankban, line 24: Nice way for setting a logger prefix!

src/charm.py
```diff
@@ -0,0 +1,161 @@
...
+        logger.debug(msg)
+        self.unit.status = WaitingStatus(msg)
+
+        # Fetch image information
+        try:
+            self.unit.status = WaitingStatus("Fetching image information ...")
+            image_info = self.image.fetch()
+        except OCIImageResourceError:
+            self.unit.status = BlockedStatus(
+                "Error fetching image information.")
```
frankban, line 108: Should we include the original error in this message?

frankban (commented):
Looks very nice with some comments and some questions.
I think the charm is easy to read and maintain. A great start and investigation.
Thanks!

.github/workflows/python-package.yml
edumucelli: Sure, I will drop it and keep only `[3.8, 3.9]`, the one that comes with Focal and the next major.

.github/workflows/python-package.yml
edumucelli: I have replaced to `ubuntu-20.04` to be explicit on what it runs on.

.jujuignore
edumucelli: Thanks, nice catch!

metadata.yaml
edumucelli: Oh, good catch! I made it up when I wrote it in the beginning. I have replaced to my e-mail, as ES operator does with authors.

README.md
edumucelli: Cool! During my week learning the juju and charms I have compiled this doc notes so that I could repeat, so then I just moved it here.

requirements.txt
edumucelli: Ok, that was automatically ordered by the requirements plugin inside the IDE, no idea what order was.

requirements.txt
edumucelli: Typo

src/charm.py
edumucelli: Good idea!

README.md
edumucelli: Sure, added.

src/charm.py
edumucelli: Neat! I will give it a try! Thanks!

src/charm.py
edumucelli: Good catch, thanks!

src/charm.py
edumucelli: Yes, it will call the `RedisClient` `close` method, unless I am missing something, maybe ...

src/charm.py
edumucelli: Good catch, thanks!

src/charm.py
edumucelli: Indeed, your suggestion is better!

src/charm.py
edumucelli: Thanks!

src/charm.py
edumucelli: Packman ate this dot :)

src/charm.py
edumucelli: Yes, good suggestion, it kind of evolved and that message stayed here. Will remove it. Thanks!

src/charm.py
edumucelli: Thanks!

src/charm.py
edumucelli: Thanks!

src/log_adapter.py
edumucelli: Thanks!

src/log_adapter.py
edumucelli: Yes, it helps a lot having the logs differentiated somehow when `juju-debug` because there a ton of them there! It was nice to have it to follow the behavior of each class.

src/charm.py
edumucelli: By briefly discussing with Thomas he told me that, for example, the `ActiveStatus` needs no message because all the administrator wants to know is if the unit is active. I think here is a bit the same. If something goes bad the trace will be in the `juju-debug` with the whole info. Here is more a quick info for the juju `dashboard` field, which is very small I think on purpose. As per doc, the `BlockedStatus` the `"operator has to manually intervene to unblock the unit and let it proceed"`, so I think all the info s/he needs is a simple line.

src/client.py
edumucelli: Sure, added, thanks!

src/charm.py
edumucelli: :+1:

src/client.py
edumucelli: I will resolve the `import` issue in a follow up PR.

src/charm.py
edumucelli: I will resolve the `import` issue in a follow up PR.

src/client.py
edumucelli: Not really needed, removed it and adapted the tests accordingly.

src/charm.py
edumucelli: Made it a regular attribute instantiated at the initialization!

src/client.py
edumucelli: Sure. That is a good point! I prefer to keep it separate because that (`client.py`) is the place where this charm will grow next, handling more stuff on the redis side, like storage, clustering. So although it is very lean today it is is meant to grow in there, not inside the `charm.py`.

src/charm.py
edumucelli: Been thinking about it, why should we sort alphabetically? It should follow [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports)
1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

src/charm.py
edumucelli: Yes, it is here.

src/charm.py
edumucelli: Sometimes it logs and defers the incoming event and once it returns directly. Kind of complicated to the gain I think. If you have a suggestion I am open.

src/charm.py
frankban: Yes I meant alphabetical order inside each group.

edumucelli: I see.

PR #4, opened 2021-02-10
Apply toms feedback

edumucelli:
This PR applies the changes from the feedback given by @mthaddon. I have 

* Moved the whole techie part of the documentation to a DEVELOPMENT.md (this also will be present in the discourse-based documentation)
* Simplified the README.md -- still lacks expanding the usage part
* Replace the upstream image with the `ubuntu/redis` image
   - This required changing the spec as the underlying image explicitly asks for empty password
* I have qualified the lack of replication/clustering and added the respective issues in this repo.

I am still not sure about the relation part. I have left a comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,112 +1,18 @@
-# redis-operator
+# Redis Operator
 
-## Description
+A Juju charm deploying and managing Redis on Kubernetes.
+
+# Overview
 
 The [Redis](https://www.redis.io/) operator provides in-memory data structure 
 store, used as a database, cache, and message broker. This repository contains a
 [Juju](https://jaas.ai/) Charm for deploying Redis on Kubernetes
 clusters.
 
-This charm does not support Redis clustering capabilities.
+This charm is in development, and it supports simple Redis topology. Although multiple
+units are allowed, nor replication nor clustering are supported for the moment. You can
+track the development in [this](https://github.com/canonical/redis-operator/issues/2) 
+and [this](https://github.com/canonical/redis-operator/issues/3) issues, respectively.
 
-## Setup, build and deploy
+# Usage
 
-A typical setup using [snaps](https://snapcraft.io/), for deployments
-to a [microk8s](https://microk8s.io/) cluster can be done using the
-following commands.
-
-Install the dependencies:
-
-    sudo snap install juju --classic
-    sudo snap install microk8s --classic
-    microk8s.enable dns storage
-    
-Create a controller named `micro` into the cloud `microk8s`:
-  
-    juju bootstrap microk8s micro
-
-In Juju, you interact with the client (the `juju` command on your local machine). 
-It connects to a controller. The controller is hosted on a cloud and controls models.
-
-    juju add-model redis-model
-
-Then you build and deploy this charm into the model you just created:
-    
-    charmcraft build
-    juju deploy ./redis.charm --resource redis-image=redis:6.0
-
-Once Redis starts up it will be running on its default port, 6379. 
-To check it you run:
-
-    juju status
-
-to discover the IP Redis is running behind. The output will have lines like:
-
-    Unit       Workload    Agent  Address       Ports     Message
-    redis/20   active      idle   10.1.168.69   6379/TCP  Pod is ready.
-
-Then, from your local machine, you can:
-
-    redis-cli -h 10.1.168.69 -p 3000
-
-Another option is to port-forward the Redis pod or service port from k8s to a local port.
-For example, forwarding a node's port
-
-    microk8s.kubectl --namespace=redis-model port-forward pod/redis-8485b69dfd-6hgvj 6379:6379
-
-However, it is simpler to forward the service port as it will abstract the node name behind it:
-
-    microk8s.kubectl --namespace=redis-model port-forward service/redis 6379:6379
-
-Then you can simply:
-
-    redis-cli
-
-From a k8s non-charmed scenario Redis will be exposed as `Service` named `redis` on the default
-port `6379`.
-
-## Developing
-
-The charm is based on the [operator framework](https://github.com/canonical/operator/). Create and activate 
-a virtualenv with the development requirements:
-
-    virtualenv -p python3 venv
-    source venv/bin/activate
-    pip install -r requirements-dev.txt
-
-In order to check the result of a modification, rebuild and upgrade the charm:
-
-    # Consider now that you are inside redis-operator directory.
-    charmcraft build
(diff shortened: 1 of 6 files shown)


src/pod_spec.py
```diff
@@ -62,6 +63,11 @@ def build_pod_spec(self) -> Dict:
 
         return spec
 
+    def _build_env_config(self):
+        return {
+            "ALLOW_EMPTY_PASSWORD": "yes"
```
edumucelli, line 69: I do not have good feelings about it, although for the csteam use case I'd be ok with it. However, I could also generate the password in the operator and attach it to a k8s opaque secret, then would pass to the relation the pass, if needed. However I think that has other implications in the code like all units must have the same password, etc. What do you guys think?

src/charm.py
```diff
@@ -153,6 +154,24 @@ def update_status(self, _):
...
+
+        Using this information a client can establish a connection with
+        Redis, for instances using the redis Python library.
+        """
+
+        if not self.unit.is_leader():
+            logger.debug("Relation changes ignored by non-leader")
+            return
+
+        event.relation.data[self.unit]['host'] = self.app.name
```
edumucelli: How this relation data is formatted normally? I have seen other using `host/port` or `uri`? As I have never created a charm before I am wondering if having only one info like `redis://pass@host:port` would also be better for the client.

edumucelli: Should this instead of advertising the redis app name, which is also is the k8s service name, be the `binding address`, e.g., 

```
def relation_changed(self, event):
   ...
   relation_id = self.model.get_relation(event.relation.name, event.relation.id)
   event.relation.data[self.unit]['host'] = self.model.get_binding(relation_id).network.bind_address
```

README.md
```diff
@@ -1,112 +1,19 @@
...
-To show the logs for all the units in the model:
-
-    microk8s.kubectl get all --namespace=redis-model
-
-To see the logs for a specific pod:
-    
-    microk8s.kubectl --namespace=redis-model logs redis-operator-0
-
-## Testing
+This charm is in development, and it supports simple Redis topology. Although multiple
```
mthaddon: "supports a simple Redis topology"?

README.md
```diff
@@ -1,112 +1,19 @@
...
+units are allowed, nor replication nor clustering are supported for the moment. You can
```
mthaddon: "replication and clustering are not supported for the moment" or "neither replication nor clustering are supported for the moment"?

README.md
```diff
@@ -1,112 +1,19 @@
...
+track the development in [this](https://github.com/canonical/redis-operator/issues/2) 
+and [this](https://github.com/canonical/redis-operator/issues/3) issues, respectively.
 
-The Python operator framework includes a very nice harness for testing
-operator behaviour without full deployment. Just `run_tests`:
+# Usage
 
-    ./run_tests
+    juju deploy redis
```
mthaddon: Until you've published the charm somewhere this isn't going to work (And will actually try to deploy a different charm currently I believe). I think you'll need to stick with "download the code, charmcraft build, deploy locally" for now.

metadata.yaml
```diff
@@ -30,8 +31,9 @@ summary: >
   of on-disk persistence, and provides high availability via Redis Sentinel and
   automatic partitioning with Redis Cluster.
 
-  This charm supports Redis 6 in Kubernetes environments, using k8s services
-  for load balancing. This charm does not support Redis clustering capabilities.
+  This charm supports Redis 5 in Kubernetes environments, using k8s services
+  for load balancing. This supports simple Redis topology. Although multiple
```
mthaddon: "supports a simple Redis topology"?

metadata.yaml
```diff
@@ -30,8 +31,9 @@ summary: >
...
+  units are allowed, nor replication nor clustering are supported for the moment.
```
mthaddon: Same comment as above about not/neither.

src/pod_spec.py
mthaddon: I think it'd be best to file an issue about this and refer to it as something you plan to fix in the future. How to do this is a similar problem as is described on https://discourse.charmhub.io/t/about-sharing-data-between-units/4165 fwiw.

tests/test_charm.py
```diff
@@ -30,11 +30,7 @@ def setUp(self):
         self.harness = Harness(RedisCharm)
         self.addCleanup(self.harness.cleanup)
         redis_resource = {
-            "registrypath": "redis:6.0",
-            # "username" and "password" are useless, but oci-resource
-            # library fetch() fails if we do not provide them ...
-            "username": "",
-            "password": ""
+            "registrypath": "redis:6.0"
```
mthaddon: Worth changing this also to match what you have in metadata.yaml?

src/pod_spec.py
edumucelli: Thanks for the link, that was exactly my doubt, I have seen one example in the MSSQL charm using the peer technique, but I was not aware about the leadership library. Good you mention it there. The issue is in https://github.com/canonical/redis-operator/issues/7

README.md
edumucelli: Good catch! Thanks!

README.md
edumucelli: Thanks for the rewording, that was what I meant :-)

metadata.yaml
edumucelli: Thanks, consistent on the error :D

metadata.yaml
edumucelli: Applied, thanks!

tests/test_charm.py
edumucelli: Yes, that is a good point, at least it won't confuse people reading it. Thanks!

README.md
edumucelli: Alright, changed accordingly.

src/charm.py
mhilton: In general I would look at what the other redis charms do, juju doesn't have any central idea of the standard for these things there is more of a de-facto standard based around interoperability.

mhilton: there is nothing stopping you putting in extra information if you think it will be useful.

mhilton (approved):
LGTM

src/charm.py
edumucelli: Alright, I saw it uses `hostname`, `port` and `password` (empty if not given). I have replace then my `host` with `hostname`.

edumucelli: The other redis charm uses the binding address. So I have done the same.

PR #5, opened 2021-02-11
No need to check state of spec before applying pod-spec-set

mthaddon:
There's no need to check the state of the spec object running set_spec - Juju will only apply this if something has changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 *.charm
 build/
 __pycache__
+.coverage
 .idea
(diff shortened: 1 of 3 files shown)


edumucelli (approved):
Alright, thanks! I got this behavior from [this part from the mongodb operator](https://github.com/canonical/mongodb-operator/blob/fdc1fe208b7539e100c959f7ae6eaaa773d17733/src/charm.py#L103-L107). Good to know that is not needed.

PR #6, opened 2021-02-11
Avoid polluting test output with logs and assert expected logging

mthaddon:
Avoid polluting test output with logs and assert expected logging in test_when_redis_fails_to_connect_then_redis_is_not_ready


edumucelli (approved):
Neat!

PR #9, opened 2021-05-11
Rename charm to redis-k8s to avoid collision with machine charm

mthaddon:
Rename charm to redis-k8s to avoid collision with machine charm


edumucelli (approved):
Thanks @mthaddon !

PR #13, opened 2021-05-12
Create and use a library for the redis relation

mthaddon:
Create and use a library for the redis relation.

This has been tested with the discourse-k8s charm to work.


edumucelli (approved):
Thanks! It seems magical to me that there is not need to 

```
self.framework.observe(self.on["datastore"].relation_changed, self.relation_changed)
```

PR #16, opened 2022-02-22
updating references for repo rename

ghost:


edumucelli (approved):

metadata.yaml
```diff
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-name: redis-k8s
+name: redis-k8s-operator
```
quiet-ranger: Is it a good idea/necessary to expose such a long name? Wouldn't it be better to just call it `redis`?  Data Platform is adopting this standard.

quiet-ranger (approved):

metadata.yaml
edumucelli: Interesting, I think this is a good idea, I wonder if that is of large adoption at Canonical.

zmraul: So, in this case the name would be redis-k8s since it's a k8s charm. For a machine charm it would be just redis. That is the convention that is being followed by charms right now. The prometheus-k8s charm has been referenced as a good baseline to follow. But I don't think any of this is really set on stone so ¯\\\_(ツ)\_/¯

zmraul: PS: I'm referring only to the metadata.yaml name, the other places look good.

edumucelli: Done

PR #17, opened 2022-03-07
Tests and CI update

zmraul:
This PR updates the remaining parts to follow the operator template.
- Migrate testing to tox
- Fix linting
- Remove old testing files
- Update documentation to reflect changes


.github/workflows/python-package.yml
```diff
@@ -1,38 +0,0 @@
-# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
```
edumucelli, line 1: Why no more Github actions? This is completely unrelated to the changes for a sidecar approach, or am I missing something here?

.flake8
```diff
@@ -1,9 +0,0 @@
-[flake8]
```
edumucelli, line 1: Why this was removed, how it is related to the sidecar change?

edumucelli:
> This PR updates the remaining parts to follow the operator template.
> 
>     * Migrate testing to tox
> 
>     * Fix linting
> 
>     * Remove old testing files
> 
>     * Update documentation to reflect changes

I am not familiar with the sidecar approach, but it seems there are plenty of unrelated changes in this PR. I'd suggest making the needed changes for the sidecar and then another PR for what is not.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          provider: microk8s
+      - name: Run integration tests
+        # set a predictable model name so it can be consumed by charm-logdump-action
+        run: tox -e integration -- --model testing
+      - name: Dump logs
+        uses: canonical/charm-logdump-action@main
+        if: failure()
+        with:
+          # TEMPLATE-TODO: Replace the application name
```
WRFitch: Since you've already done this, you can remove this comment

tests/test_charm.py
```diff
@@ -15,15 +15,20 @@
...
 
-from ops.model import ActiveStatus, Container, MaintenanceStatus, UnknownStatus, WaitingStatus
-from ops.pebble import ServiceInfo, ConnectionError
-from ops.testing import Harness
-
 from charm import RedisK8sCharm
-from charms.redis_k8s.v0.redis import RedisProvides
 
 
 class TestCharm(TestCase):
```
WRFitch, line 34: Could you please move this file to tests/integration/test_charm.py?

zmraul:
Hi Eduardo! I understand the title can lead to confusion, I can change it to just `template update` or something else if that is the case. As you said, the changes here don't really affect anything sidecar related.

This PR updates the testing and CI to follow the [operator-template](https://github.com/canonical/operator-template).

These and following changes will target the `update` branch, which is related, but not limited to the sidecar update. This way we can work on features/fixes while keeping `main` as it is right now. Does this sound good? or should I do it some other way.

WRFitch (changes requested):
Needs a couple of small changes, but otherwise lgtm

edumucelli:
> Hi Eduardo! I understand the title can lead to confusion, I can change it to just `template update` or something else if that is the case. As you said, the changes here don't really affect anything sidecar related.
> 
> This PR updates the testing and CI to follow the [operator-template](https://github.com/canonical/operator-template).
> 
> These and following changes will target the `update` branch, which is related, but not limited to the sidecar update. This way we can work on features/fixes while keeping `main` as it is right now. Does this sound good? or should I do it some other way.

I see, I was not aware there was a template, and that tests and ci were guided by it as well. Sure, it is good, go ahead.

edumucelli (approved):

tests/test_charm.py
zmraul: Yes! Thanks for catching that.
PS: It will be tests/**unit**/test_charm.py

WRFitch (approved):

tests/test_charm.py
WRFitch: Yes, it is - sorry, thanks for catching that!

src/charm.py
```diff
@@ -14,24 +14,27 @@
...
 REDIS_PORT = 6379
-WAITING_MESSAGE = 'Waiting for Redis...'
+WAITING_MESSAGE = "Waiting for Redis..."
 
 logger = logging.getLogger(__name__)
 
 
 class RedisK8sCharm(CharmBase):
+    """Charm the service."""
+
```
pengale, line 41: This is an opportunity to talk a little bit about what this charm does. Obviously, this class is the core charm for the service. But what does it currently do, specifically? Is there anything that someone editing this class might need to know?

src/charm.py
```diff
@@ -44,6 +47,7 @@ def __init__(self, *args):
         self.framework.observe(self.on.check_service_action, self.check_service)
 
     def config_changed(self, event):
+        """Handle config_changed event."""
```
pengale: Again, this could use a descriptive docstring. Something that would work better might be:

"Possibly update the container.

On config changed, check to see if we need to update the container contents. If we do, update and restart the redi service."

src/charm.py
```diff
@@ -83,10 +85,12 @@ def config_changed(self, event):
         self._redis_check()
 
     def update_status(self, event):
+        """Handle update_status event."""
```
pengale: Same here, as far as docs go.

"On update status, check the container."

src/charm.py
```diff
@@ -83,10 +85,12 @@ def config_changed(self, event):
...
         logger.info("Beginning update_status")
         self._redis_check()
 
     def check_service(self, event):
+        """Handle for check_service action."""
```
pengale: And here: "When requested, check service."

pengale (commented):
Overall, this looks good, and represents a nice update to the current charm template.

The doc strings are very generic, though, and this is a great opportunity to make them a bit more specific.

zmraul:
> Overall, this looks good, and represents a nice update to the current charm template.
> 
> The doc strings are very generic, though, and this is a great opportunity to make them a bit more specific.

Hi @pengale, I agree with all you said. I've added a bit of explanation to each docstring. However, I've been working on a follow up PR that will refactor several parts of the code, to prepare for new features like replication. So these docstrings might get changed pretty soon :)

pengale (approved):
Nice changes to the docs. +1 :-)

charmcraft.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - build-on:
+      - name: "ubuntu"
+        channel: "20.04"
+    run-on:
+      - name: "ubuntu"
+        channel: "20.04"
+parts:
+  charm:
+    build-packages:
+      - git
```
marceloneppel: Do you need `git` for some dependency? I saw that the repo no longer uses git dependencies (like `git+https://github.com/juju-solutions/resource-oci-image#egg=oci-image`).

marceloneppel (approved):
Great PR @zmraul! I just left one question about the `git` build package.

charmcraft.yaml
zmraul: I don't, thanks for letting me know! Some leftovers from the template :)

PR #20, opened 2022-03-22
DPE-116 Add initial admin password

zmraul:
### Problem

This PR address JIRA ticket [DPE-116](https://warthogs.atlassian.net/browse/DPE-116).
By default,  databases should be deployed with an initial admin password, increasing the security of the system.
 
### Solution

A randomly generated password, 16 characters long, is generated on the leader unit, if no previous password existed.

The peer relation databag is used to store the password. Although the current status of the charm only deploys Redis as a standalone replica, with this solution the new changes will make use of the existing peer relation storage system. Using the peer databag will also help in maintaining the same password during scaling up/down of the application.

### Context

The password is being injected into the database using the [ubuntu/redis](https://hub.docker.com/r/ubuntu/redis) environment parameters. The entrypoint script to the container updates the Redis config file to reflect the password change. Because of this, there is no need (yet) to use a config template.

### Release Notes

The Pebble layer logic has been refactored to ease future updates to the code.

- `requirements.txt`: `ops` has been updated to 1.3.0 to use of `container.can_connect()` method
- `src/charm.py`
   - `config_changed()`: this method has been simplified, creating two more with specific logic: `_redis_layer()` that returns an updated Pebble layer, and `_update_layer()` tasked with updating Pebble if needed.
   - The peer relation databag now stores a password.
   - Some helper charm methods have been added; `_generate_password()`, `_get_password()` and a property for the peer relation `_peers()`

### PS
Integration tests will be fixed on a next PR when an action to get the current password is added.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -54,3 +54,7 @@ resources:
     type: oci-image
     description: ubuntu lts docker image for redis
     upstream: ubuntu/redis
+
+peers:
+  redis-peers:
+    interface: redis-peers
--- a/src/charm.py
+++ b/src/charm.py
@@ -33,6 +33,8 @@ REDIS_PORT = 6379
 WAITING_MESSAGE = "Waiting for Redis..."
 CONFIG_PATH = "/etc/redis/"
 PEER = "redis-peers"
+PEER_PASSWORD_KEY = "redis-password"
+
 
 logger = logging.getLogger(__name__)
 
@@ -71,13 +73,13 @@ class RedisK8sCharm(CharmBase):
         will be stored on the peer relation databag.
         """
         peer_data = self._peers.data[self.app]
-        redis_password = peer_data.get("redis-password", None)
+        redis_password = peer_data.get(PEER_PASSWORD_KEY, None)
 
         if redis_password is None:
-            self._peers.data[self.app]["redis-password"] = self._generate_password()
+            self._peers.data[self.app][PEER_PASSWORD_KEY] = self._generate_password()
 
             # TODO: remove this once the action to retrieve the password is implemented
-            logger.info("REDIS PASSWORD: {}".format(peer_data.get("redis-password", None)))
+            logger.info("REDIS PASSWORD: {}".format(peer_data.get(PEER_PASSWORD_KEY, None)))
 
     def _config_changed(self, event: EventBase) -> None:
         """Handle config_changed event.
@@ -208,7 +210,7 @@ class RedisK8sCharm(CharmBase):
             String with the password
         """
         data = self._peers.data[self.app]
-        return data.get("redis-password", "")
+        return data.get(PEER_PASSWORD_KEY, "")
 
 
 if __name__ == "__main__":  # pragma: nocover
--- a/src/charm.py
+++ b/src/charm.py
@@ -161,6 +161,7 @@ class RedisK8sCharm(CharmBase):
         Returns:
             An `ops.model.Relation` object representing the peer relation.
         """
+        return self.model.get_relation(PEER)
 
 
 if __name__ == "__main__":  # pragma: nocover
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,7 @@ from charms.redis_k8s.v0.redis import RedisProvides
 from ops.charm import CharmBase
 from ops.main import main
 from ops.model import ActiveStatus, WaitingStatus, Relation
-from ops.pebble import Layer
+from ops.pebble import Layer, ConnectionError
 from redis import Redis
 from redis.exceptions import RedisError
 
@@ -44,7 +44,6 @@ class RedisK8sCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
 
-        self._container = self.unit.get_container("redis")
         self.redis_provides = RedisProvides(self, port=REDIS_PORT)
 
         self.framework.observe(self.on.redis_pebble_ready, self._redis_pebble_ready)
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -130,6 +168,48 @@ def _redis_check(self):
...
+        return password
+
+    def _get_password(self) -> str:
+        """Get the current admin password for Redis.
+
+        Returns:
+            String with the password
+        """
+        data = self._peers.data[self.app]
+        return data.get("redis-password", "")
```
edumucelli: nit: `"redis-password"` to become a constant, close to where `PEER` is. `PEER` is used once and has a constant, `"redis-password"` is used at least 3 times.

edumucelli (approved):

src/charm.py
zmraul: Yeah, it makes sense. I'll update the code :)

metadata.yaml
```diff
@@ -54,3 +54,7 @@ resources:
     type: oci-image
     description: ubuntu lts docker image for redis
     upstream: ubuntu/redis
+
+peers:
+  redis-peers:
+    interface: redis-peers
```
marcoppenheimer, line 60: I would probably aim to be more specific here with the name of the relation - or is this generic for Cluster + non-Cluster?

src/charm.py
```diff
@@ -44,79 +51,112 @@ def __init__(self, *args):
...
+
+        If no password exists, a new one will be created for accessing Redis. This password
+        will be stored on the peer relation databag.
+        """
+        peer_data = self._peers.data[self.app]
+        redis_password = peer_data.get(PEER_PASSWORD_KEY, None)
+
+        if redis_password is None:
+            self._peers.data[self.app][PEER_PASSWORD_KEY] = self._generate_password()
+
```
marcoppenheimer: I would probably b64 encode the password sitting the databag, 'security through obscurity'. In reality it's not doing anything to protect it, but is vanishingly safer.

src/charm.py
```diff
@@ -44,79 +51,112 @@ def __init__(self, *args):
...
+    def _leader_elected(self, _) -> None:
+        """Handle the leader_elected event.
+
+        If no password exists, a new one will be created for accessing Redis. This password
+        will be stored on the peer relation databag.
+        """
+        peer_data = self._peers.data[self.app]
+        redis_password = peer_data.get(PEER_PASSWORD_KEY, None)
+
+        if redis_password is None:
```
marcoppenheimer: I think this `get(_, None)` is nice and explicit, but you might also achieve the same by defaulting `None` to `self._generate_password()` and skipping the `if`.

To make that clearer for a user, you could add a `logging.debug` in the `_generate_password()` call to say `'Generating default password'` or something to that effect.

src/charm.py
```diff
@@ -44,79 +51,112 @@ def __init__(self, *args):
...
+            },
+        }
+        return Layer(layer_config)
+
+    def _redis_check(self) -> None:
+        """Checks is the Redis database is active."""
         try:
-            redis = Redis()
+            redis = Redis(password=self._get_password())
             info = redis.info("server")
```
marcoppenheimer, line 163: If I'm assuming correctly, you're looking here to confirm that you have a healthy Redis connection. I'd change this line to:
```
response = redis.ping()
if response == 'PONG':
    self.unit.status = ActiveStatus()
else:
    stuff()
```
I can't remember exactly how, but I think `ping` does a few things to check for healthy connection, and is used internally for the connection handlers for most of the Python client libraries for destroying and retrying connections, so it's good for parity to use that also.

marcoppenheimer (approved):
I like the interface for keeping the password, and makes sense to not have the action yet.
Great work, minor comments!

src/charm.py
```diff
@@ -130,6 +170,48 @@ def _redis_check(self):
...
+        return password
+
+    def _get_password(self) -> str:
+        """Get the current admin password for Redis.
+
+        Returns:
+            String with the password
+        """
+        data = self._peers.data[self.app]
+        return data.get(PEER_PASSWORD_KEY, "")
```
paulomach, line 216: I'm not sure if defaulting to empty string `""` is the best here. For instance, if you need to test the password existance, by using this function it will be required to test it like `if _get_password() == "":` instead of just boolean testing the function, i.e. `if _get_password():`.

paulomach (approved):
I second Marc's comments plus a minor one. All in all, looks good

src/charm.py
zmraul: Yeah, I added the _get_password() afterwards and didn't realize I can just use it here. But I will need the check, as the generation of the password should only be done once, the first leader election. I'll update this.

metadata.yaml
zmraul: Yes! I'll probably need to change this when adding the cluster/replication if more relations are needed.

src/charm.py
zmraul: An empty string will evaluate to false:
```python
>>> a = ""
>>> bool(a)
False
```
So, in that sense it shouldn't be a problem. But I'm unsure if it's a bit too obscure, and it would be just better to return None.

shayancanonical (approved):
lgtm, but support the already existing review comments

src/charm.py
zmraul: Calling the info command is done to update the Juju status information with the Redis version that is running. Now that I look at the docstring, I see that I messed it up :) The method checks that the database is active as a side effect really. The main purpose I guess is to check Redis _version_. I can extract this, or make a couple of helpers, one for updating this information and another one for doing a proper health check.

src/charm.py
paulomach: :exploding_head: Didn't knew that !

zmraul: heh, python is weird :)

marceloneppel (approved):

PR #21, opened 2022-03-25
DPE-117 Add password action

zmraul:
### Problem

This PR address JIRA ticket [DPE-117](https://warthogs.atlassian.net/browse/DPE-117).
When deploying the database with a default password, an action should be provided to retrieve that password.
 
### Solution

The charm will implement an action to return the initial password.

### Context

This charm stores the password on the peer relation databag.

### Release Notes

- `actions.yaml`: new action created; `get-initial-admin-password`
- `src/charm.py`
   - `_get_password_action()`:  Handler for the action. Will create a result with the key `redis-password`.
- `tests/integration/test_charm.py`
   - `test_application_is_up()`: now uses the password
   - `test_database_with_no_password()`: check that the database is not accessible without a password
   - `test_same_password_after_scaling()`: scale down and up the application to test that the password remains the same

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -10,6 +10,7 @@ import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
 from redis import Redis
+from redis.exceptions import AuthenticationError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_charm.py
```diff
@@ -45,8 +46,71 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    """Check that the database cannot be accessed without a password."""
+    status = await ops_test.model.get_status()  # noqa: F821
+    address = status["applications"][APP_NAME]["units"][f"{APP_NAME}/0"]["address"]
+
     cli = Redis(address)
+    try:
+        cli.ping()
+        assert False, "Client can connect without password"
+    except AuthenticationError:
+        assert True
```
WRFitch: You can rewrite this to be more terse: 
```python
with pytest.raises(AuthenticationError):
    cli.ping()
```

tests/integration/test_charm.py
```diff
@@ -45,8 +46,71 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    # Use action to get admin password
+    logger.info("calling action to retrieve password")
+    action = await ops_test.model.units.get(f"{APP_NAME}/0").run_action(
+        "get-initial-admin-password"
+    )
+    password = (await action.wait()).results["redis-password"]
+
+    cli = Redis(address, password=password)
+
+    assert cli.ping()
```
WRFitch: I understand that we need this action to access the password and therefore create the Redis object, but it makes it less clear that this test tests database access, since it will fail if the `get-initial-admin-password` action fails. Could this action be tested separately, or at least given an explicit assertion in this test to ensure the reason for possible failures is communicated to the developer?

WRFitch (changes requested):
Looks good, some notes about your tests

tests/integration/test_charm.py
```diff
@@ -45,8 +46,71 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    logger.info("scaling charm %s to 1 units", APP_NAME)
+    await ops_test.model.applications[APP_NAME].scale(scale=1)
+    await ops_test.model.block_until(lambda: len(ops_test.model.applications[APP_NAME].units) > 0)
+
+    # Use action to get admin password after scaling
+    logger.info("calling action to retrieve password")
+    action = await ops_test.model.units.get(f"{APP_NAME}/0").run_action(
+        "get-initial-admin-password"
+    )
+    after_pw = (await action.wait()).results["redis-password"]
```
marceloneppel: You can create a helper method to encapsulate the logic of calling the action to avoid duplication.

marceloneppel (commented):
LGTM Raúl! One note in the integration tests.

shayancanonical (approved):
LGTM, but support the already existing comments on the PR

tests/integration/test_charm.py
zmraul: Thanks for the suggestion! Added :)

tests/integration/test_charm.py
zmraul: In this case, the password will be either a correct string or an empty string. If the password is not correct, the test will fail regardless with an AuthenticationError. Maybe a debug log with the retrieved password?

tests/integration/test_charm.py
zmraul: Yes! I extracted the code into a helper method.

tests/integration/test_charm.py
WRFitch: My point was more that they're separate functions and should therefore be tested separately, similar to how unit tests work. However, the test and the action are both simple enough that it's not a big deal, especially if you add some extra logging.

WRFitch (approved):

marceloneppel (approved):

PR #27, opened 2022-04-20
DPE-214 user defined tls certificates

zmraul:
## Issue
This PR addresses Jira ticket [DPE-214](https://warthogs.atlassian.net/browse/DPE-214).
Databases should have support for TLS encryption. This charm now allows a user to add certificates and enable the usage of TLS on Redis.

## Solution
The charm will implement a solution to store TLS certificates as resources and enable them through a config option.

Three new resources have been added to the charm. A config option has been created as well to enable/disable the usage of the certificates. The expected way to enable them would be:
```shell
juju deploy redis-k8s_ubuntu-20.04-amd64.charm --resource redis-image=ubuntu/redis

# Generate test certificates using the script provided
./tests/gen-test-certs.sh

# Attach the files as resources
juju attach-resource redis-k8s cert-file=tests/tls/redis.cert
juju attach-resource redis-k8s key-file=tests/tls/redis.key
juju attach-resource redis-k8s ca-cert-file=tests/tls/ca.cert

# Enable TLS
juju config redis-k8s enable-tls=true
```

The connection to the redis-server doesn't enforce mutual TLS, so a client using the `redis-cli` needs to connect with the appropriate flags (`ssl=true` and `ssl_ca_certs=/path/to/ca.cert`).

## Context
The `juju attach-resource` command triggers a `upgrade_charm` event. At this moment, only the certificates are handled on the event. When the event triggers, the certificates are copied to the Redis container. This is done so `redis-server` is initialized with the paths to the certificate files. Since the event doesn't specify what resource is changing, a list of all found certificates is copied again into the workload container every time the event triggers.

## Release Notes
- `config.yaml`: added `enable-tls` option, by default set to false
- `metadata.yaml`: added three resources, needed for each certificate file
- `src/redis_client.py`: helper methods to create a Redis client connection with the specified parameters
- `src/charm.py`:
  - `_upgrade_charm()`: will store certificates on the mounted folder of the unit's container
  - `_redis_extra_flags()`: creates the extra parameters based on the charm's config (right now only for certificates)
  - `_config_changed()`: update to use the new optional flags
  - `_redis_check()`: now generates the redis client using the helpers on `src/redis_client.py`
  - `_certificates()`: property with the paths to the resources or None when no resource has been added
  - `_store_certificates()`: will copy the found resource into the container storage path
  - `_retrieve_resource()`: will retrieve the path given a resource name
- `tests/gen-test-certs.sh`: script that creates TLS certificate files used for testing
- `tests/helpers.py`: place where useful variables/methods are located

### Integration
Minor changes to the deployment to take resources into account.

- `test_blocked_if_no_certificates`: tests that enabling tls without providing certificates first results in a Blocked status
- `test_enable_tls`: attaches the certificates and enables tls, expecting an Active status and connecting to the redis server afterwards.
- Some helper methods have been added at the bottom to ease testing

### Unit
Minor changes to the tests to add the extra flags variable.

- ~~`test_blocked_status_on_wrong_resource_name`: this cannot be tested directly through the harness.add_resource(), as it fails before going into the charm code. So a direct call to the `RedisK8sCharm._retrieve_resource()` method has been implemented instead~~ removed
- `test_attach_resource`: tests that when a resource is attached, is copied
- `test_blocked_on_enable_tls_with_no_certificates`: tests a blocked status on config_changed with no resources attached
- `test_active_on_enable_tls_with_certificates`: tests that the charm is active if certificates have been provided

## Final notes
Right now the certificates themselves are not being checked, so redis-server can break if bad certificates are attached. This will result on a Waiting status with a message of `Waiting for redis...`. Further checks will be implemented to give more information about this situation.


paulomach:
Didn't review it yet, but the summary is stellar!

config.yaml
```diff
@@ -0,0 +1,11 @@
...
+# See LICENSE file for licensing details.
+
+options:
+  enable-tls:
+    default: false
+    type: boolean
+    description: >
+      Enabling TLS for the standard port of redis-server.
+      This option should be enabled after uploading the
+      corresponding TLS files as resources.
```
paulomach: nit: add new line

metadata.yaml
```diff
@@ -37,23 +37,37 @@ summary: >
   units are allowed, replication and clustering are not supported for the moment.
 maintainers:
   - Eduardo Mucelli R. Oliveira <eduardo.mucelli@canonical.com>
```
paulomach, line 39: Add `Raul Zamora`

src/charm.py
```diff
@@ -54,45 +57,52 @@ def __init__(self, *args):
...
     def _config_changed(self, event: EventBase) -> None:
         """Handle config_changed event.
 
         Updates the Pebble layer if needed. Finally, checks the redis service
         updating the unit status with the result.
         """
+        # Check that certificates for TLS exist
+        if self.config["enable-tls"] and None in self._certificates:
+            logger.warning("Not enough certificates found for TLS")
+            self.unit.status = BlockedStatus("No certificates found")
```
paulomach: Lifecycle question here: if instead of blocking, use a waiting status+defer, would it be possible to attach the certificates an reprocess it? 
It's a bit more gentle to the end user.

src/charm.py
```diff
@@ -225,6 +268,39 @@ def _get_password(self) -> str:
...
         return data.get(PEER_PASSWORD_KEY, "")
 
+    def _store_certificates(self) -> None:
+        """Copy the TLS certificates to the redis container."""
+        # Get a list of valid paths
+        cert_paths = list(filter(None, self._certificates))
+
+        # Copy the files from the resources location to the redis container.
+        for cert_path in cert_paths:
+            shutil.copy(cert_path, self._storage_path)
```
paulomach: I'm curious about how this works. It seems counter intuitive to me.

src/charm.py
```diff
@@ -225,6 +268,39 @@ def _get_password(self) -> str:
...
+    def _retrieve_resource(self, resource: str) -> Optional[str]:
+        """Check that the resource exists and return it.
+
+        Returns:
+            string with the path of the resource or None
+        """
+        try:
+            # Fetch the resource path
+            return self.model.resources.fetch(resource)
+        except ModelError as e:
```
paulomach: In the same vibe as previous comment about blocking status, some comments above.
Additionally, my feeling is that helper functions shouldn't trigger status changes, but throw errors and let the calling event handler execute the reasoning about status, deferring and so on.

tests/helpers.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
```
paulomach, line 1: I like this approach, single point of truth for some literals!

tox.ini
```diff
@@ -73,4 +73,8 @@ deps =
     pytest-operator
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
\ No newline at end of file
+    # Generate TLS certificates to use during tests
+    sh "./tests/gen-test-certs.sh"
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
+whitelist_externals =
+    sh
```
paulomach: nit: add new line

paulomach (commented):
Some minor changes, but mostly clarifications.

src/charm.py
zmraul: In this case, the blocking status will be set because the user changed the config `enable-tls` to true, but didn't attach all the files. I think that it needs to be blocked, as it's a situation that requires user interaction no matter what. The user will either set the config back to false, or it will add the files to fix this problem.

src/charm.py
zmraul: I agree, it creates fleeting Blocked statuses (trying to query all files) that will be fixed right after. I'll remove the status changes altogether and handle them somewhere else.

src/charm.py
zmraul: Looking at the code comment, I'll change it to `# Check that certificates exist if TLS is enabled`

paulomach: Your approach is correct

shayancanonical (approved):

tests/gen-test-certs.sh
```diff
@@ -0,0 +1,23 @@
...
+    -subj '/O=Redis Test/CN=Server' | \
+    openssl x509 \
+        -req -sha256 \
+        -CA tests/tls/ca.crt \
+        -CAkey tests/tls/ca.key \
+        -CAserial tests/tls/ca.txt \
+        -CAcreateserial \
+        -days 365 \
+        -out tests/tls/redis.crt
+openssl dhparam -out tests/tls/redis.dh 2048
```
WRFitch, line 23: Since this part takes some time, is it worth including an early return in this script if the required certs already exist?

src/redis_client.py
```diff
@@ -0,0 +1,26 @@
...
+
+    Returns:
+        Redis class with the connection
+    """
+    redis = Redis(password=password)
+    if ssl:
+        ca_path = f"{storage_path}/ca.crt"
+        redis = Redis(password=password, ssl=ssl, ssl_ca_certs=ca_path)
+
+    return redis
```
WRFitch, line 26: nit: please add a blank line

tests/gen-test-certs.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
WRFitch, line 1: You may want to add the output of this file into your .gitignore

tox.ini
```diff
@@ -73,4 +73,8 @@ deps =
     pytest-operator
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
\ No newline at end of file
+    # Generate TLS certificates to use during tests
+    sh "./tests/gen-test-certs.sh"
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
```
WRFitch: Just a suggestion: I've found it useful to add `--durations=0` to integration tests, so I can see if there's any tests that take a while.

tests/integration/test_charm.py
```diff
@@ -109,25 +112,91 @@ async def test_same_password_after_scaling(ops_test: OpsTest):
...
+        return
+
+    logger.info(f"Attaching resource: attach-resource {APP_NAME} {rsc_name}={rsc_path}")
+    await ops_test.juju("attach-resource", APP_NAME, f"{rsc_name}={rsc_path}")
+
+
+async def change_config(ops_test: OpsTest, conf_name: str, conf_value: Any) -> None:
+    """Use the `juju config` command to modify a config option."""
+    logger.info(f"Changing config option: {conf_name}={conf_value}")
+    await ops_test.juju("config", APP_NAME, f"{conf_name}={conf_value}")
```
WRFitch: This is a simpler way to change config: 
```python
await ops_test.model.applications[APP_NAME].set_config({name:value})
```
You could update this to pass in a dict so you can set multiple values simultaneously, minimising `await`s.

tests/integration/test_charm.py
```diff
@@ -109,25 +112,91 @@ async def test_same_password_after_scaling(ops_test: OpsTest):
...
+
+async def attach_resource(ops_test: OpsTest, rsc_name: str, rsc_path: str) -> None:
+    """Use the `juju attach-resource` command to add resources."""
+    # NOTE: Is this check needed for a functional test?
+    if rsc_name not in METADATA["resources"]:
+        logger.error(f"Invalid resource name: {rsc_name}")
+        return
+
+    logger.info(f"Attaching resource: attach-resource {APP_NAME} {rsc_name}={rsc_path}")
+    await ops_test.juju("attach-resource", APP_NAME, f"{rsc_name}={rsc_path}")
```
WRFitch, line 197: This may be a more elegant method of changing config, depending on your definition of elegant: 
```python
await ops_test.model.applications[APP_NAME].attach_resource(rsc_name, rsc_path, rsc_file_object)
```
It should also remove the need for the above check.

tests/integration/test_charm.py
```diff
@@ -109,25 +112,91 @@ async def test_same_password_after_scaling(ops_test: OpsTest):
...
+
+    Will enable TLS without providing certificates. This should result
+    on a Blocked status.
+    """
+    await change_config(ops_test, "enable-tls", "true")
+
+    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="blocked", timeout=1000)
+
+    logger.info("trying to check for blocked status")
+    assert ops_test.model.applications[APP_NAME].units[0].workload_status == "blocked"
```
WRFitch, line 131: Exiting tests in a status that isn't active-idle means the tests aren't idempotent, and therefore their order can't be easily changed. I think it's worth it to return the charm back to its original state with the following lines:

```python
await change_config(ops_test, "enable-tls", "true")
await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)
```

src/charm.py
zmraul: I'll change it to `container.push()` as it is a better approach.

src/redis_client.py
zmraul: I think there is already one?

tests/gen-test-certs.sh
zmraul: Definitely, I'll update

tests/integration/test_charm.py
zmraul: Yeah! I wasn't sure on where to 'reset' things. When the tests grow more it might be worth to have a cleanup method that is run at the end of some tests.

tests/gen-test-certs.sh
zmraul: I think that the script should only create the certs, as if you want to create new ones it won't run.
Maybe I should update `tox.ini` commands to remove the files afterwards. The problem with the script taking some time persists though, so I'm not really happy with that solution :/

tox.ini
zmraul: I'll try that :)

tests/integration/test_charm.py
zmraul: Agree, seems like a good idea.

tests/integration/test_charm.py
zmraul: I went for the `ops_test.juju()` because your option was creating some problems, but can't find the issue where this was discussed. I'll test changing the code

tests/gen-test-certs.sh
WRFitch: A simpler option might be to just commit these as static test files, since they're only used for testing. This way, you could also include testing for invalid certs if you wanted.

tests/integration/test_charm.py
WRFitch: Resetting to a clean state after every test is valuable because otherwise your tests all implicitly rely on each other, so debugging one test really means debugging every test up to that point. I don't think the state has to be totally pristine though, you can get away with leaving files around. 

A cleanup method might be handy. You can define setup and teardown hooks in pytest that may make this easier.

src/charm.py
```diff
@@ -206,6 +239,16 @@ def _peers(self) -> Optional[Relation]:
...
 
+    @property
+    def _certificates(self) -> List[Optional[Path]]:
+        """Paths of the certificate files.
+
+        Returns:
+            A list with the paths of the certificates or None where no path can be found
+        """
+        resources = ["cert-file", "key-file", "ca-cert-file"]
+        return [self._retrieve_resource(res) for res in resources]
```
marceloneppel, line 250: Maybe you can use something like below to not need to filter `None` values outside this function:

```python
return [path for res in resources if (path := self._retrieve_resource(res)) is not None]
```

marceloneppel (approved):
LGTM!

src/charm.py
zmraul: I thought about this, but I'm checking if None is on the list on L101. I could check for the size of the list instead, but I'm not sure what is the best approach.

marceloneppel: I think both the None or the size of the list checks are fine.

PR #29, opened 2022-05-04
Update charm to support `redis` relation interface

zmraul:
This PR updates the charm so if the `redis` relation shows up, the charm will disable the password authentication. Some deprecation warnings have been added to the logs.

### DIFF
--- a/lib/charms/redis_k8s/v0/redis.py
+++ b/lib/charms/redis_k8s/v0/redis.py
@@ -44,7 +44,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version.
-LIBPATCH = 2
+LIBPATCH = 3
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/redis_k8s/v0/redis.py
+++ b/lib/charms/redis_k8s/v0/redis.py
@@ -93,8 +93,15 @@ class RedisProvides(Object):
         """A class implementing the redis provides relation."""
         super().__init__(charm, "redis")
         self.framework.observe(charm.on.redis_relation_changed, self._on_relation_changed)
+        self.framework.observe(charm.on.redis_relation_created, self._on_relation_created)
         self._port = port
+        self.charm = charm
 
+    def _on_relation_created(self, _):
+        """Handle the relation created event."""
+        self.charm._peers.data[self.charm.app]["enable-password"] = "false"
+        self.charm._update_layer()
+    
     def _on_relation_changed(self, event):
         """Handle the relation changed event."""
         if not self.model.unit.is_leader():
--- a/lib/charms/redis_k8s/v0/redis.py
+++ b/lib/charms/redis_k8s/v0/redis.py
@@ -99,6 +99,8 @@ class RedisProvides(Object):
 
     def _on_relation_created(self, _):
         """Handle the relation created event."""
+        # TODO: Update warning to point to the new interface once it is created
+        logger.warning("DEPRECATION WARNING - `redis` interface is a legacy interface.")
         self.charm._peers.data[self.charm.app]["enable-password"] = "false"
         self.charm._update_layer()
     
--- a/lib/charms/redis_k8s/v0/redis.py
+++ b/lib/charms/redis_k8s/v0/redis.py
@@ -101,8 +101,9 @@ class RedisProvides(Object):
         """Handle the relation created event."""
         # TODO: Update warning to point to the new interface once it is created
         logger.warning("DEPRECATION WARNING - `redis` interface is a legacy interface.")
-        self.charm._peers.data[self.charm.app]["enable-password"] = "false"
-        self.charm._update_layer()
+        if self.charm.unit.is_leader():
+            self.charm._peers.data[self.charm.app]["enable-password"] = "false"
+            self.charm._update_layer()
     
     def _on_relation_changed(self, event):
         """Handle the relation changed event."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -143,9 +143,7 @@ class RedisK8sCharm(CharmBase):
         # have a password.
         new_layer = self._redis_layer()
         if self._peers.data[self.app].get("enable-password", "true") == "false":
-            logger.warning(
-                "DEPRECATION WARNING - charm configured with password off"
-            )
+            logger.warning("DEPRECATION WARNING - charm configured with password off")
             env = new_layer.services["redis"].environment
             env["ALLOW_EMPTY_PASSWORD"] = "yes"
             if "REDIS_PASSWORD" in env:
--- a/src/charm.py
+++ b/src/charm.py
@@ -143,7 +143,9 @@ class RedisK8sCharm(CharmBase):
         # have a password.
         new_layer = self._redis_layer()
         if self._peers.data[self.app].get("enable-password", "true") == "false":
-            logger.warning("DEPRECATION WARNING - charm configured with password off")
+            logger.warning(
+                "DEPRECATION WARNING - password off, this will be removed on later versions"
+            )
             env = new_layer.services["redis"].environment
             env["ALLOW_EMPTY_PASSWORD"] = "yes"
             if "REDIS_PASSWORD" in env:
--- a/src/charm.py
+++ b/src/charm.py
@@ -138,6 +138,18 @@ class RedisK8sCharm(CharmBase):
         # Create the new config layer
         new_layer = self._redis_layer()
 
+        # NOTE: This block is to allow the lecagy `redis` relation interface to work
+        # with charms still using it. Charms using the relation don't expect Redis to
+        # have a password.
+        new_layer = self._redis_layer()
+        if self._peers.data[self.app].get("enable-password", "true") == "false":
+            logger.warning(
+                "DEPRECATION WARNING - password toggle off will not be supported in the future"
+            )
+            env = new_layer.services["redis"].environment
+            env["ALLOW_EMPTY_PASSWORD"] = "yes"
+            if "REDIS_PASSWORD" in env: del env["REDIS_PASSWORD"]
+
         # Update the Pebble configuration Layer
         if current_layer.services != new_layer.services:
             logger.debug("About to add_layer with layer_config:\n{}".format(new_layer))
@@ -194,7 +206,10 @@ class RedisK8sCharm(CharmBase):
         """Checks is the Redis database is active."""
         try:
             redis = redis_client(
-                self._get_password(), self.config["enable-tls"], self._storage_path
+                self._peers.data[self.app].get("enable-password", "true"),
+                self._get_password(),
+                self.config["enable-tls"],
+                self._storage_path,
             )
             info = redis.info("server")
             version = info["redis_version"]
--- a/src/charm.py
+++ b/src/charm.py
@@ -144,7 +144,7 @@ class RedisK8sCharm(CharmBase):
         new_layer = self._redis_layer()
         if self._peers.data[self.app].get("enable-password", "true") == "false":
             logger.warning(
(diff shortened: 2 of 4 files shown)


lib/charms/redis_k8s/v0/redis.py
```diff
@@ -93,8 +93,15 @@ def __init__(self, charm, port):
         """A class implementing the redis provides relation."""
         super().__init__(charm, "redis")
         self.framework.observe(charm.on.redis_relation_changed, self._on_relation_changed)
+        self.framework.observe(charm.on.redis_relation_created, self._on_relation_created)
         self._port = port
+        self.charm = charm
 
+    def _on_relation_created(self, _):
+        """Handle the relation created event."""
+        self.charm._peers.data[self.charm.app]["enable-password"] = "false"
```
marcoppenheimer: Just a thought, but I *believe* (this could just be opinion though) that when using boolean flags like this, it's good to phrase it as an action against what is thought of as the default.
Made up examples could be: `disable_security`,`enable_experimental_features`, `dont_send_logs` etc.

In this case, I think the flag should be named `disable-password`, if we want to encourage having it be there by default.

src/charm.py
```diff
@@ -138,6 +138,18 @@ def _update_layer(self) -> None:
         # Create the new config layer
         new_layer = self._redis_layer()
 
+        # NOTE: This block is to allow the lecagy `redis` relation interface to work
+        # with charms still using it. Charms using the relation don't expect Redis to
+        # have a password.
+        new_layer = self._redis_layer()
+        if self._peers.data[self.app].get("enable-password", "true") == "false":
+            logger.warning(
+                "DEPRECATION WARNING - password toggle off will not be supported in the future"
```
marcoppenheimer:
```suggestion
                "DEPRECATION WARNING - disabling passwords with the XYZ-password option will not be supported in the future"
```

src/charm.py
```diff
@@ -138,6 +138,18 @@ def _update_layer(self) -> None:
         # Create the new config layer
         new_layer = self._redis_layer()
 
+        # NOTE: This block is to allow the lecagy `redis` relation interface to work
+        # with charms still using it. Charms using the relation don't expect Redis to
+        # have a password.
```
marcoppenheimer: I really like this comment 📝

src/redis_client.py
```diff
@@ -7,18 +7,24 @@
...
 
     Returns:
         Redis class with the connection
     """
-    redis = Redis(password=password)
+    redis = Redis()
+
+    # NOTE: This check will become deprecated in the future
+    if password_enabled == "true":
+        redis = Redis(password=password)
```
marcoppenheimer: I'm not sure we want to override the `redis` reference here, it might be safer to `if password_enabled: else` here, returning only a single connection.

It's probably fine as it is, but not certain how the connection pool gets GC'd so best to be safe imo.

marcoppenheimer (commented):
Overall it looks great! I have some design questions (which we may have discussed before, so forgive me if I forgot).

As I understand, for a normal user, before any peer relation is established:

1. `_update_layer` - Redis Unit initialises with a password, set through env-vars/pebble
2. `_redis_check` - We confirm we are able to access the live redis service. `enable-password` will be non-existent (for now), so defaults to 'true'
3. `redis_client` - We establish a redis connection object using the password (because `password_enabled` is 'true'

Now, if a new user deploys this charm, and uses the relation in `lib/charms/redis_k8s/v0/redis.py`, just by using it, they no longer have passwords at all, with no way to change it (because it's hardcoded in source)


Assuming I understood that right, I now have actual questions. 
1. How are we ensuring that any prospective 'new' user will not be using this relation, despite it being in the source-code and looks good?
2. Is it intentional that any user (new or old) has no ability to 'toggle' passwords themselves? USE RELATION - No passwords for you. DON'T USE RELATION - Passwords only, no way out. It's perfectly fine, I wonder if a `config.yaml` option to `disable_passwords`, defaulting to 'false' would help avoid hard-coding `get(__, bool)` within the source where it's a bit obscure.

The reason for both those questions, is I think we can be more explicit with THIS relation specifically, with some unique identifier (this isn't a good suggestion, but say it adds a `legacy` string into the relation data). Then, `when 'legacy' or config.option: set env_vars and reapply layer`. Or by naming the relation 'Legacy' and looking for that string in the event. I dunno.
New users can turn off passwords if they want (though defaults are obvious and clear in `config.yaml`), the old relation is made obvious somehow, but doesn't break backwards. 

Most of what I said is probably rubbish, but what do you think?

src/charm.py
WRFitch: I like it too, but you misspelt "legacy" ;)

lib/charms/redis_k8s/v0/redis.py
zmraul: In this case, the default would be true, as in `enable-password = true`. But is only used to be changed within the relation to false.
I think here, on the relation code, what you are saying makes more sense and in the charm code the other way around makes more sense. Is `.get("disable-password", "false")` more intuitive within the charm context?

zmraul:
> As I understand, for a normal user, before any peer relation is established:
> 
> 1. `_update_layer` - Redis Unit initialises with a password, set through env-vars/pebble
> 
> 2. `_redis_check` - We confirm we are able to access the live redis service. `enable-password` will be non-existent (for now), so defaults to 'true'
> 
> 3. `redis_client` - We establish a redis connection object using the password (because `password_enabled` is 'true'

This is the intended workflow, yes!

> Now, if a new user deploys this charm, and uses the relation in `lib/charms/redis_k8s/v0/redis.py`, just by using it, they no longer have passwords at all, with no way to change it (because it's hardcoded in source)

Correct, the current `redis` relation interface doesn't store the password on the relation databag. Not only that, some of the charms connecting are not expecting a password at all when connecting to the Redis charm.

> 1. How are we ensuring that any prospective 'new' user will not be using this relation, despite it being in the source-code and looks good?

Hopefully, good documentation and log warnings will take care of that. But ultimately we can't force users not to use that relation. I guess same with any other framework  :)
In the future this relation will be removed entirely, this is done to give time to other charm teams to make the changes they need to use the new features of the redis charm.

> 2. Is it intentional that any user (new or old) has no ability to 'toggle' passwords themselves? USE RELATION - No passwords for you. DON'T USE RELATION - Passwords only, no way out. It's perfectly fine, I wonder if a `config.yaml` option to `disable_passwords`, defaulting to 'false' would help avoid hard-coding `get(__, bool)` within the source where it's a bit obscure.

I think there is not going to be an option to toggle off passwords at the moment (outside of giving support for relations like this one). Maybe this will change in the future, but currently the only possible deployment should be the _secure_ one.

> The reason for both those questions, is I think we can be more explicit with THIS relation specifically, with some unique identifier (this isn't a good suggestion, but say it adds a `legacy` string into the relation data). Then, `when 'legacy' or config.option: set env_vars and reapply layer`. Or by naming the relation 'Legacy' and looking for that string in the event. I dunno. New users can turn off passwords if they want (though defaults are obvious and clear in `config.yaml`), the old relation is made obvious somehow, but doesn't break backwards.

I think `redis` relation will become the legacy one. Moving forward, new relations for all databases will share a common naming pattern, so a new relation will have to be created. I can definitely add some extra documentation at the top of the library, but supporting the `redis` relation _as is_ means that the `RedisRequires()` part of the library stays the same.

> Most of what I said is probably rubbish, but what do you think?

Definitely not, always helps to have an outside view into what we take for granted :)

lib/charms/redis_k8s/v0/redis.py
marcoppenheimer: This makes sense given the context from your comment below, i.e this isn't a 'user' flag, it's a hard-coded one in source. As such, it's great as is.

marcoppenheimer:
OK all your comments make sense, and I agree with it. 
I'm apprehensive about having 'There's this code here but don't use it', but hopefully with more obvious `This is the right relation` info it should be ok, I'll approve when it's out of draft! :)

jnsgruk:
It may be work for another PR, but we need to add an integration test that deploys/relates another charm to this one on the `redis` interface to make sure we don't introduce any regressions as we iterate on the newer interface.

zmraul:
> It may be work for another PR, but we need to add an integration test that deploys/relates another charm to this one on the `redis` interface to make sure we don't introduce any regressions as we iterate on the newer interface.

I have tested the relation locally with the discourse bundle. But deploying the three charms involved is a somewhat big time sink. Is there a good way to do integration tests in situations like these? Or just to deploy everything and wait for it to settle

jnsgruk:
Yeh I mean they likely will take a while in CI, but your unit tests should still allow you to iterate fairly quickly locally? The other option is to find something smaller than uses redis... @mthaddon might have an idea here?

lib/charms/redis_k8s/v0/redis.py
```diff
@@ -93,8 +93,17 @@ def __init__(self, charm, port):
...
         self.framework.observe(charm.on.redis_relation_changed, self._on_relation_changed)
+        self.framework.observe(charm.on.redis_relation_created, self._on_relation_created)
         self._port = port
+        self.charm = charm
 
+    def _on_relation_created(self, _):
+        """Handle the relation created event."""
+        # TODO: Update warning to point to the new interface once it is created
+        logger.warning("DEPRECATION WARNING - `redis` interface is a legacy interface.")
+        self.charm._peers.data[self.charm.app]["enable-password"] = "false"
```
marceloneppel: I guess you need to do this operation only in the redis leader unit to avoid some errors related to the other units trying to set this data in the application data bag.

marceloneppel (commented):
It looks great. I left only one comment related to the relation data that's being set.

lib/charms/redis_k8s/v0/redis.py
zmraul: You are right, thanks for catching this. The problem now is what happens when the leader unit doesn't get the event the first one. All units should reconfigure by calling `update_layer()` but if a secondary unit runs `relation_created` and tries to update the Pebble layer, password will still be there because the peer data didn't change.
I can modify the peer data only on the leader, and after that `defer` if the data hasn't been modified for each unit?

mthaddon:
I'm not aware of a smaller bundle than discourse & postgres & redis I'm afraid. Note that using the edge channel of discourse may make that faster as it's using sidecar, and the image is being pulled from the charmhub registry rather than dockerhub.

zmraul:
Moving it out of draft, as the integration tests for the relation will come on a future PR using something similar to [database relation charm](https://github.com/zmraul/db-relations-k8s)

src/charm.py
```diff
@@ -138,6 +138,17 @@ def _update_layer(self) -> None:
         # Create the new config layer
         new_layer = self._redis_layer()
 
+        # NOTE: This block is to allow the legacy `redis` relation interface to work
+        # with charms still using it. Charms using the relation don't expect Redis to
+        # have a password.
+        new_layer = self._redis_layer()
+        if self._peers.data[self.app].get("enable-password", "true") == "false":
+            logger.warning("DEPRECATION WARNING - charm configured with password off")
```
marcoppenheimer: For a user, they'll likely only see the logs. I'd add a sentence after this saying 'This will be removed in later versions' or something

marcoppenheimer (approved):
That's a really good test.

lib/charms/redis_k8s/v0/redis.py
zmraul: @marceloneppel Turns out the application databag can only be modified by the leader, so I was thinking this in a wrong way. I have added a check for leadership (otherwise `self.charm._peers.data[self.charm.app]["enable-password"] = "false"` would fail on any other non-leader unit) and handled the relation `redis_peers_relation_changed` event on the charm code.

WRFitch (approved):

marceloneppel (approved):

PR #33, opened 2022-06-08
DPE-401 Replication basics

zmraul:
# Issue

This PR addresses Jira ticket [DPE-401](https://warthogs.atlassian.net/browse/DPE-401).
Create the replication logic to deploy Redis with one leader and **n** replicas. All new units will replicate the leader one.

# Solution

The charm implements a solution to enable replication by default. The leader unit will be the master, and all other units will become replicas. Replication could be disabled by changing a specific config option.

The fastest way to check if replication is working would be as follows:
```
# deploy with several units
juju deploy redis-k8s_ubuntu-20.04-amd64.charm --resource redis-image=ubuntu/redis -n 3

# connect to the redis instance using either a python client or redis-cli (apt install redis-tools)
redis-cli <unit-ip> -a <password>
> SET "myKey" "value"

# setting a value on the leader unit will be replicated on the rest
> GET "myKey"
b"value"
```

Currently, this implementation allows deployment of several units (`juju deploy . . . -n 3`) and scale up (`juju add-unit redis-k8s`)

# Context

Replication is enabled by passing command-line arguments to redis-server at startup. Since the commands need to specify the master ip to replicate, the leader unit will add it's own hostname to the peer databag, so other units now the name of the leader. Then, redis-server is called with the extra arguments `--replicaof <leader-hostname> <port> --masterauth <password>`.

Some checks for the `enable-password` have been scattered around, so legacy `redis` relation can be used with replication.

# Release Notes

- ~~`config.yaml`: added `enable-replication` option, by default set to **true**~~(removed)
- `src/redis_client.py`: removed, can be handled directly on charm.py
- `src/literals.py`: extract literals to their own file (future PR will need them on their own file)
- `src/charm.py`:
  - `_redis_extra_flags()`: creates the extra parameters for replication for non leader units
  - `_current_master`: property with the hostname of the current redis master
  - `_valid_app_databag()`: checks if the application data has been set. This is useful when non-leader units are ready before the leader has populated the application databag
  - `_get_pod_hostname()`: transforms a pod name into a pod domain hostname
  - `_get_redis_client()`: now uses the default values of Redis constructor to generate a client

# Testing

### Integration
Now the number of units deployed initially has been parametrized, defaults to 2. Some tests have been improved to leave the model in the same state as the initial.

Some marks have been added, so specific tests can be run on development environments:

`tox -e integration -- -vv --model <mymodel> --no-deploy -m replication_tests`

- `test_replication`: after scaling up, check that the leader-set keys on redis are the same on the new replicas
- ~~`test_disable_replication`: tests that all units are masters after disabling replication~~(removed)
- A helper method has been added at the bottom to ease testing

### Unit

- `test_non_leader_unit_as_replica`: checks that the pebble layer for non leader units has replication flags added.

# Final notes

Automatic failover cannot be guaranteed with this implementation, Sentinel will be added on a following PR to ensure HA requirements are met.


src/charm.py
```diff
@@ -291,14 +338,18 @@ def _generate_password(self) -> str:
...
         Returns:
             String with the password
         """
         data = self._peers.data[self.app]
-        return data.get(PEER_PASSWORD_KEY, "")
+        # NOTE: (DEPRECATE) When using redis legacy relation, no password is used
+        if data.get("enable-password", "true") == "false":
+            return None
+
+        return data.get(PEER_PASSWORD_KEY, None)
```
marceloneppel: I guess the  `None` keyword is the default value, so you can remove it.

tests/integration/test_charm.py
```diff
@@ -10,11 +10,12 @@
 from redis import Redis
 from redis.exceptions import AuthenticationError
 
-from tests.helpers import APP_NAME, METADATA, TLS_RESOURCES
+from tests.helpers import APP_NAME, METADATA, NUM_UNITS, TLS_RESOURCES
 
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.skip_if_deployed
```
marceloneppel, line 18: Good to know about this :+1:

tests/integration/test_charm.py
```diff
@@ -171,6 +184,68 @@ async def test_enable_tls(ops_test: OpsTest):
     cli = Redis(address, password=password, ssl=True, ssl_ca_certs=TLS_RESOURCES["ca-cert-file"])
     assert cli.ping()
 
+    await change_config(ops_test, {"enable-tls": "false"})
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", raise_on_blocked=False, timeout=1000
+    )
```
marceloneppel, line 193: Maybe it's worth making another call using the Redis client with TLS turned off to make sure it's been turned off? Maybe you can ignore as it's being tested in other tests and this block of code seems to be related to only reset the state of the deployment.

config.yaml
```diff
@@ -9,3 +9,9 @@ options:
       Enabling TLS for the standard port of redis-server.
       This option should be enabled after uploading the
       corresponding TLS files as resources.
+  enable-replication:
```
paulomach: I wonder if is there's any case when a user deploying more then 1 unit don't want replication. Is there a difference in deploying a single unit with this flag as `false` from a deployment with this set to `true`?

src/charm.py
```diff
@@ -217,17 +234,27 @@ def _redis_extra_flags(self) -> str:
...
+            # relation interface is being used.
+            if self._peers.data[self.app].get("enable-password", "true") == "true":
+                extra_flags += [f"--masterauth {self._get_password()}"]
+
+        logger.debug("Extra flags: {}".format(extra_flags))
+
         return " ".join(extra_flags)
 
     def _redis_check(self) -> None:
         """Checks is the Redis database is active."""
```
paulomach:
```suggestion
        """Checks if the Redis database is active."""
```

src/charm.py
```diff
@@ -321,9 +373,31 @@ def _retrieve_resource(self, resource: str) -> Optional[Path]:
             # Fetch the resource path
             return self.model.resources.fetch(resource)
         except (ModelError, NameError) as e:
-            logger.error(e)
+            logger.warning(e)
             return None
 
+    def _get_pod_hostname(self, pod_name: str) -> str:
+        """Creates the pod hostname from its name."""
+        return f"{pod_name}.{self._name}-endpoints.{self._namespace}.svc.cluster.local"
```
paulomach: On the recent talk on libraries, we saw that observability team is using `socket.getfqdn`  function for this. Did not try it though.

src/charm.py
```diff
@@ -321,9 +373,31 @@ def _retrieve_resource(self, resource: str) -> Optional[Path]:
...
+
+    def _get_redis_client(self, hostname="localhost") -> Redis:
```
paulomach: This is a good case for [context manager](https://docs.python.org/3.9/library/contextlib.html#contextlib.contextmanager)

paulomach (commented):
Left some comments all around.
Also, unrelated to the PR, I've notices in the charm code that you are setting `app.status` directly. I'm not sure that's necessary.

marceloneppel (approved):
LGTM! The integrations tests worked with `--agent-version 2.9.29` (in the bootstrap) from the link that you shared on MM.

config.yaml
zmraul: There is no difference. I'm not too sure if there is a need for stopping replication. I think not, having the option to deploy the same charm with different alias would already cover that use case I guess.

src/charm.py
zmraul: I'll give it a try

tests/integration/test_charm.py
zmraul: Yeah, this part is just to reset the model. I can add the check though

config.yaml
zmraul: I removed the option to stop replication. If it's needed down the line is not a huge effort to bring it back.

shayancanonical (approved):
LGTM!

PR #34, opened 2022-06-13
DPE-419 Add Sentinel container

zmraul:
# Issue

This PR addresses Jira ticket [DPE-419](https://warthogs.atlassian.net/browse/DPE-419).
One of the steps towards HA, is to provide a reliable automatic failover process. 

A following PR will add logic to keep the leader hostname updated using sentinel, so scale operations after a failover are successful. Similarly, quorum is set to 1 at the moment, as specific actions to handle scale up/down will be covered in the future.

This PR should be addressed after merging #33.

## Solution

To enable automatic failover, Sentinel has been added to a sidecar container. After initial replication, Sentinel will take over the tasks of monitoring failures on redis-server and trigger failovers.

The fastest way to check if sentinel is working would be as follows:
```
# deploy with several units
juju deploy redis-k8s_ubuntu-20.04-amd64.charm --resource redis-image=ubuntu/redis -n 3

# connect to the sentinel instance using either a python client or redis-cli (apt install redis-tools)
redis-cli <unit-ip> -p 26379
> INFO "sentinel"
```
The output should show the total number of sentinels match the units deployed.

## Context

Sentinel is deployed on a sidecar container. The charm will implement a handler for the Sentinel pebble layer and a template to pass the configuration to sentinel.

## Release Notes

- `src/literals.py`: extended with pertinent values
- `src/sentinel.py`:
  - `_sentinel_pebble_ready()`: creates the pebble layer for sentinel and adds the template to the container

## Testing

Further tests will be implemented on a specific PR

### Integration

- `test_sentinels_expected`: check that the number of sentinels matches the number of deployed units


tests/integration/test_charm.py
```diff
@@ -224,6 +224,20 @@ async def test_replication(ops_test: OpsTest):
     leader_client.close()
 
 
+@pytest.mark.replication_tests
+async def test_sentinels_expected(ops_test: OpsTest):
+    """Test sentinel connection and expected number of sentinels."""
+    # TODO extract number getter to a funct?
```
WRFitch: I think this is worthwhile, since it repeats what you're doing on line 215. It should also make your code a little more readable and less fragile :)

WRFitch (commented):
LGTM once you follow up on your TODO comments :)

marcoppenheimer (approved):
Fab PR 👍🏽

marceloneppel (approved):
Great PR! LGTM.

PR #36, opened 2022-06-23
Cleanup

zmraul:
This PR is a mix of bugfixes and improvements to update several aspects of the charm.

* The charm now uses hostnames everywhere, instead of IPs.
    * Earlier versions of redis don't support hostname announcement/resolution so the container has been changed to use Redis 6.2. At the moment, `ubuntu/redis:6.2` doesn't ship with the correct version, so `redis:6.2` has been used instead.
* Now the Pebble layer calls `redis-server` directly, without passing through the `entry_point.sh` script.
    * As a side consequence, the environment variables are no longer needed. The `REDIS_PASSWORD` has been moved to the `extra_flags` method
    * All unit tests have been updated to the new schema
* Pebble layer has added `redis` as unix user:group
    * `container.push()` now specifies user:group and permissions
* Sentinel now uses a password (different from the redis server)
* Sentinel now uses `app.name` as master name, to avoid clashes on _autodiscovery_ of replicated instances. (This could happen if two redis applications are deployed with different aliases)
* Sentinel has a template condition to check for master password
* Downgrade bootstrap agent to `2.9.29` to avoid [scale-down bug](https://bugs.launchpad.net/juju/+bug/1977582) on integration tests

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -210,9 +210,6 @@ class RedisK8sCharm(CharmBase):
                     "user": "redis",
                     "group": "redis",
                     "startup": "enabled",
-                    "environment": {
-                        "ALLOW_EMPTY_PASSWORD": "yes",
-                    }
                 }
             },
         }
--- a/src/charm.py
+++ b/src/charm.py
@@ -25,7 +25,9 @@ from literals import (
     LEADER_HOST_KEY,
     PEER,
     PEER_PASSWORD_KEY,
+    REDIS_CONFIG_PATH,
     REDIS_PORT,
+    SENTINEL_PASSWORD_KEY,
     WAITING_MESSAGE,
 )
 from sentinel import Sentinel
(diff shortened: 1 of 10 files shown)


marcoppenheimer (approved):

src/charm.py
```diff
@@ -227,9 +226,12 @@ def _redis_extra_flags(self) -> str:
                 f"--tls-ca-cert-file {self._storage_path}/ca.crt",
             ]
 
-        # Non leader units will be replicas of the leader unit
-        if not self.unit.is_leader():
-            extra_flags += [f"--replicaof {self._current_master} {REDIS_PORT}"]
+        # Check that current unit is master
+        if self.current_master != self.unit_pod_hostname:
```
MiaAltieri, line 233: Question: Why is it necessary to check if the current unit is the master? (Sorry if this is a basic question, I am new to sentinel and such :) )

MiaAltieri (approved):
LGTM, just get your integration tests passing :D

src/charm.py
zmraul: If the pebble layer is triggering, maybe because of a pod restart/reschedule, I need to point all the units to the master to enable replication. Since any unit could be the current replication master, I check beforehand so I don't pass unnecessary commands.

MiaAltieri: Great explaination! thanks Raul :)

PR #37, opened 2022-07-01
DPE-464 Scale-up logic

zmraul:
# Issue

This PR addresses Jira ticket [DPE-464](https://warthogs.atlassian.net/browse/DPE-464).
One of the steps towards HA, is to provide a reliable automatic failover process. 

This PR adds logic to keep the leader hostname updated using sentinel, so scale operations after a failover are successful. Similarly, quorum is propagated to all unit when scaling the charm.

## Solution

Sentinel will take over the tasks of monitoring failures on redis-server and trigger failovers. If a failover occurs, the next scaling operation or relation related operation should be aware of the changes to configure correctly. At the start of the `_peer_relation_handler`, there are checks to query who is the replication master. Additionally, quorum is set to all units to a new value when units are joining.

## Context

Sentinel is deployed on a sidecar container. Connecting to Sentinel will help the charm understand the current replication status.

## Release Notes

- `scr/sentinel.py`:
  - `_sentinel_pebble_ready()`: now the pebble layer will replan, as it is only called on start sequence or pod rescheduling.
  - `get_master_info()`: connect to sentinel and query current master info
  - `expected_quorum`: calculate current quorum based on juju application scale.
  - `in_majority`: check that own sentinel can reach quorum.
  - `sentinel_client()`: generate a sentinel client.
- `src/charm.py`:
  - `_check_master()`: check that the master info stored on the charm matches real replication master.
  - `_update_application_master()`: store the current master on the application databag.
  - `_k8s_hostname()`: create a DNS hostname from a unit name.
  - `_update_quorum()`: update the quorum of all application units.

## Testing

Further tests will be implemented on a specific PR

### Integration

- `test_scale_up_replication_after_failover`: Add some keys to the database. Force a failover, then scale the application up. After this check that new units are replicating correctly and have their keys.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,7 +18,7 @@ from ops.framework import EventBase
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, ModelError, Relation, WaitingStatus
 from ops.pebble import Layer
-from redis import Redis
+from redis import ConnectionError, Redis, TimeoutError
 from redis.exceptions import RedisError
 
 from literals import (
@@ -27,6 +27,7 @@ from literals import (
     PEER_PASSWORD_KEY,
     REDIS_PORT,
     SENTINEL_PASSWORD_KEY,
+    SOCKET_TIMEOUT,
     WAITING_MESSAGE,
 )
 from sentinel import Sentinel
@@ -69,13 +70,18 @@ class RedisK8sCharm(CharmBase):
 
         self._storage_path = self.meta.storages["database"].location
 
-    def _redis_pebble_ready(self, _) -> None:
+    def _redis_pebble_ready(self, event) -> None:
         """Handle the pebble_ready event.
 
         Updates the Pebble layer if needed.
         """
         self._store_certificates()
         self._update_layer()
+
+        # update_layer will set a Waiting status if Pebble is not ready
+        if self.unit.status != ActiveStatus():
+            event.defer()
+            return
 
     def _upgrade_charm(self, _) -> None:
         """Handle the upgrade_charm event.
@@ -132,24 +138,43 @@ class RedisK8sCharm(CharmBase):
         On update status, check the container.
         """
         logger.info("Beginning update_status")
+        if self.unit.is_leader():
+            self._update_application_master()
         self._redis_check()
 
     def _peer_relation_handler(self, event):
         """Handle relation for joining units."""
-        self._update_layer()
+        if not self._check_master():
+            if self.unit.is_leader():
+                # Update who the current master is
+                self._update_application_master()
+            self._update_layer()
 
-        # update layer should leave the unit in active status
-        if self.unit.status != ActiveStatus():
+        if not (self.unit.is_leader() and event.unit):
+            return
+
+        if not self.sentinel.in_majority:
+            self.unit.status = WaitingStatus("Waiting for majority")
             event.defer()
             return
 
-    def _on_redis_relation_created(self, _):
+        # Update quorum for all sentinels
+        self._update_quorum()
+
+    def _on_redis_relation_created(self, event):
         """Handle the relation created event."""
         # TODO: Update warning to point to the new interface once it is created
         logger.warning("DEPRECATION WARNING - `redis` interface is a legacy interface.")
-        if self.unit.is_leader():
-            self._peers.data[self.app]["enable-password"] = "false"
-            self._update_layer()
+        if not self.unit.is_leader():
+            return
+
+        self._peers.data[self.app]["enable-password"] = "false"
+        self._update_layer()
+
+        # update_layer will set a Waiting status if Pebble is not ready
+        if self.unit.status != ActiveStatus():
+            event.defer()
+            return
 
     def _update_layer(self) -> None:
         """Update the Pebble layer.
@@ -175,7 +200,6 @@ class RedisK8sCharm(CharmBase):
 
         # Update the Pebble configuration Layer
         if current_layer.services != new_layer.services:
-            logger.debug("About to add_layer with layer_config:\n{}".format(new_layer))
             container.add_layer("redis", new_layer, combine=True)
             logger.info("Added updated layer 'redis' to Pebble plan")
             container.restart("redis")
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -69,14 +70,19 @@ def __init__(self, *args):
...
+    def _redis_pebble_ready(self, event) -> None:
         """Handle the pebble_ready event.
 
         Updates the Pebble layer if needed.
         """
         self._store_certificates()
         self._update_layer()
 
+        # update_layer will set a Waiting status if Pebble is not ready
+        if self.unit.status != ActiveStatus():
```
WRFitch: I think this will evaluate to `False` if ActiveStatus has a status message - consider the following instead:
```suggestion
if isinstance(self.unit.status, ActiveStatus):
```

WRFitch: You may also want to log why this event is being deferred

src/charm.py
```diff
@@ -132,24 +138,43 @@ def _update_status(self, _) -> None:
...
-            self._peers.data[self.app]["enable-password"] = "false"
-            self._update_layer()
+        if not self.unit.is_leader():
+            return
+
+        self._peers.data[self.app]["enable-password"] = "false"
+        self._update_layer()
+
+        # update_layer will set a Waiting status if Pebble is not ready
+        if self.unit.status != ActiveStatus():
```
WRFitch: See above comment about using `isinstance()`

src/sentinel.py
```diff
@@ -135,3 +130,59 @@ def _copy_file(self, path: str, rendered: str, container: str) -> None:
...
+            try:
+                # get sentinel info about the master
+                master_info = sentinel.execute_command(f"SENTINEL MASTER {self.charm._name}")
+            except (ConnectionError, TimeoutError) as e:
+                logger.error("Error when connecting to sentinel: {}".format(e))
+
+        # NOTE: master info from redis comes like a list:
+        # ['key1', 'value1', 'key2', 'value2', ...]
+        # this creates a dictionary in a more readable form.
+        master_info = dict(zip(master_info[::2], master_info[1::2]))
```
WRFitch: If `master_info` isn't set by `sentinel.execute_command`, at this point you'll be trying to index `None` and getting an error. I have a suggestion for a replacement: 

```python
with self.sentinel_client(host) as sentinel:
    try:
        # get sentinel info about the master
        master_info = sentinel.execute_command(f"SENTINEL MASTER {self.charm._name}")

        # NOTE: master info from redis comes like a list:
        # ['key1', 'value1', 'key2', 'value2', ...]
        # this creates a dictionary in a more readable form.
        return dict(zip(master_info[::2], master_info[1::2]))

    except (ConnectionError, TimeoutError) as e:
        logger.error("Error when connecting to sentinel: {}".format(e))

return None # or {}
# or raise ConnectionError, depending on how bad it would be to not connect to sentinel
```

src/sentinel.py
```diff
@@ -135,3 +130,59 @@ def _copy_file(self, path: str, rendered: str, container: str) -> None:
...
+                if response.startswith("OK"):
+                    logger.info("Own sentinel instance can reach quorum")
+                    majority = True
+            except RedisError as e:
+                logger.warning("No quorum can be reached: {}".format(e))
+
+        return majority
+
+    @contextmanager
+    def sentinel_client(self, hostname="localhost", timeout=SOCKET_TIMEOUT) -> Redis:
```
WRFitch, line 172: Missing args in docstring

WRFitch (commented):
Looks good, I have a couple of suggestions

src/charm.py
```diff
@@ -132,24 +138,43 @@ def _update_status(self, _) -> None:
...
+            self._update_application_master()
         self._redis_check()
 
     def _peer_relation_handler(self, event):
         """Handle relation for joining units."""
-        self._update_layer()
+        if not self._check_master():
+            if self.unit.is_leader():
+                # Update who the current master is
+                self._update_application_master()
```
marcoppenheimer, line 150: Here.

src/sentinel.py
```diff
@@ -135,3 +130,59 @@ def _copy_file(self, path: str, rendered: str, container: str) -> None:
             user="redis",
             group="redis",
         )
+
+    def get_master_info(self, host="localhost") -> Optional[dict]:
```
marcoppenheimer, line 134: Is there a scenario where master might not be set while it gets called?
E.g during a deferred pebble ready, but peers are joining? 

It might be better to handle a failed master and defer/retry to give election time to complete, not sure how often this happens though.

(see `Here` comment for where the issue might arise.)

marcoppenheimer: Maybe return a boolean? 

False if failed, and handle the failure in `src/charm.py`? Not sure.

marcoppenheimer (approved):

src/charm.py
zmraul: Yes! You are right

src/sentinel.py
zmraul: Good catch!

src/sentinel.py
zmraul: Hm, let me think a bit about this, because the logic on the _peer_relation_handler method is still a bit fuzzy. 

It's overkill at the moment though, because it's triggering for all relation_changed events. Even if there is some small moment where master is moving to a new redis server, some relation_changed event down the line should pick up the change and start the reconfiguration process. It works, but it's not a very good design. 

I believe the edge case you are referring to is what would happen if a failover occurs exactly while the last `relation_changed` event triggers.

src/sentinel.py
```diff
@@ -111,7 +106,7 @@ def _render_sentinel_config_file(self) -> None:
             sentinel_port=SENTINEL_PORT,
             redis_master=self.charm.current_master,
             redis_port=REDIS_PORT,
-            quorum=1,
+            quorum=self.expected_quorum,
```
marceloneppel, line 109: Is this updated every time in all sentinel units? Can it cause a problem if it's not updated correctly on them all?

marceloneppel (approved):
LGTM!

src/sentinel.py
zmraul: At the moment this will only trigger when the unit is first being deployed or if k8s reschedules the pod. Other than this, quorum changes will be propagated using `_update_quorum` on `charm.py`

marceloneppel: Cool! Thanks for the explanation, Raúl!

src/sentinel.py
marcoppenheimer: Yeah, I think I'm fine for not hyper-bulletproofing it now. 
TBF I'm worried about most of my work when we do concurrent stuff (kill leader during scale up and scale down at the same time, or something)

WRFitch (approved):

PR #38, opened 2022-07-07
Auto release to charmhub

paulomach:
Auto release to charmhub latest/edge github workflow.
Charmhub access key added in repository secrets.

*NOTE*: The tests in the `release.yaml` a bit different from the one in `python-package.yaml` and come from the operator template repository. This should not be a issue though.


jnsgruk (approved):

zmraul:
I think this should target the `update` branch

paulomach:
> I think this should target the `update` branch

Actually, this action will trigger only on push to `main`, so it just make sense to merge to it.
It can wait here while the work on `update` it's not finished.

zmraul (approved):

PR #39, opened 2022-07-15
DPE-485 Scale down

zmraul:
# Issue

This PR addresses Jira ticket [DPE-485](https://warthogs.atlassian.net/browse/DPE-485).
Logic for scaling-down has been added.

## Solution

During scale-down, new quorum will be propagated and a reset command will be sent to all sentinels, so they remove information form departing `redis-server` and `sentinel`.
- In the event of one of the departing units being the current replication master, a manual failover will be triggered, so a new master is elected.

## Context

Some changes have been made to the initial command. The initial replication master needs replica flags as well. When a failover occurs, the old master will need those flags so it can become a secondary successfully.

## Release Notes

- `charm.py`
   - `_peer_relation_departing()` now handles scale-down logic
   - `_leader_elected()` added logic in the case juju leader departs and it was redis master
   - reworked sentinel command helpers
   - redis pebble layer bugs after failover fixed 
   - added `_sentinel_failover()` to trigger a manual failover
   - `_is_failover_finished()`
- Added custom exceptions
- Fixed directory creation for sentinel
- Updated sentinel template so failovers timeout are faster

## Testing

Unit tests using the pebble layer have been fixed to take changes into account.

### Unit

- Test application data updated after a failover happens.
- Test manual failover when departing unit is replication master.

### Integration

- Added scale down on departing redis master test.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -298,7 +298,7 @@ class RedisK8sCharm(CharmBase):
             logger.warning(
                 "DEPRECATION WARNING - password off, this will be removed on later versions"
             )
-            extra_flags = ["--bind 0.0.0.0"]
+            extra_flags = ["--bind 0.0.0.0", f"--replica-announce-ip {self.unit_pod_hostname}"]
 
         if self.config["enable-tls"]:
             extra_flags += [
--- a/src/charm.py
+++ b/src/charm.py
@@ -544,7 +544,7 @@ class RedisK8sCharm(CharmBase):
 
     def _sentinel_failover(self, departing_unit_name: str) -> None:
         """Try to failover the current master.
-        
+
         This method should only be called from juju leader, to avoid more than one
         sentinel sending failovers concurrently.
         """
--- a/src/charm.py
+++ b/src/charm.py
@@ -59,7 +59,7 @@ class RedisK8sCharm(CharmBase):
         self.framework.observe(self.on.update_status, self._update_status)
 
         self.framework.observe(self.on.redis_relation_created, self._on_redis_relation_created)
-        self.framework.observe(self.on[PEER].relation_changed, self._peer_relation_handler)
+        self.framework.observe(self.on[PEER].relation_changed, self._peer_relation_changed)
         self.framework.observe(self.on[PEER].relation_departed, self._peer_relation_departed)
 
         self.framework.observe(self.on.check_service_action, self.check_service)
@@ -113,8 +113,8 @@ class RedisK8sCharm(CharmBase):
             except Exception:
                 raise
 
-            # logger.warning("Resetting sentinel")
-            # self._reset_sentinel()
+            logger.warning("Resetting sentinel")
+            self._reset_sentinel()
 
         if not self._get_password():
             logger.info("Creating password for application")
@@ -155,7 +155,7 @@ class RedisK8sCharm(CharmBase):
             self._update_application_master()
         self._redis_check()
 
-    def _peer_relation_handler(self, event):
+    def _peer_relation_changed(self, event):
         """Handle relation for joining units."""
         if not self._check_master():
             if self.unit.is_leader():
@@ -197,8 +197,8 @@ class RedisK8sCharm(CharmBase):
             event.defer()
             return
 
-        # logger.warning("Resetting sentinel")
-        # self._reset_sentinel()
+        logger.warning("Resetting sentinel")
+        self._reset_sentinel()
 
         self.unit.status = ActiveStatus()
 
@@ -280,6 +280,7 @@ class RedisK8sCharm(CharmBase):
             f"--requirepass {self._get_password()}",
             "--bind 0.0.0.0",
             f"--masterauth {self._get_password()}",
+            f"--replica-announce-ip {self.unit_pod_hostname}",
         ]
 
         if self._peers.data[self.app].get("enable-password", "true") == "false":
@@ -300,10 +301,7 @@ class RedisK8sCharm(CharmBase):
 
         # Check that current unit is master
         if self.current_master != self.unit_pod_hostname:
-            extra_flags += [
-                f"--replicaof {self.current_master} {REDIS_PORT}",
-                f"--replica-announce-ip {self.unit_pod_hostname}",
-            ]
+            extra_flags += [f"--replicaof {self.current_master} {REDIS_PORT}"]
 
             if self.config["enable-tls"]:
                 extra_flags += ["--tls-replication yes"]
@@ -562,20 +560,21 @@ class RedisK8sCharm(CharmBase):
             raise Exception
 
     def _update_quorum(self) -> None:
-        """Iterate over the units on the relation, and update their quorum.
-
-        Connect to all Sentinels deployed to update the quorum.
-        """
+        """Connect to all Sentinels deployed to update the quorum."""
         command = f"SENTINEL SET {self._name} quorum {self.sentinel.expected_quorum}"
         self._broadcast_sentinel_command(command)
 
     def _reset_sentinel(self):
-        """Reset sentinel to process changes during scale-down operations mainly."""
+        """Reset sentinel to process changes and remove unreachable servers/sentinels."""
         command = f"SENTINEL RESET {self._name}"
         self._broadcast_sentinel_command(command)
 
     def _broadcast_sentinel_command(self, command: str) -> None:
-        """Broadcast a command to all sentinel instances."""
+        """Broadcast a command to all sentinel instances.
+
+        Args:
+            command: string with the command to broadcast to all sentinels
+        """
         hostnames = [self._k8s_hostname(unit.name) for unit in self._peers.units]
         # Add the own unit
         hostnames.append(self.unit_pod_hostname)
--- a/src/charm.py
+++ b/src/charm.py
@@ -534,7 +534,7 @@ class RedisK8sCharm(CharmBase):
 
     def _sentinel_failover(self, departing_unit_name: str) -> None:
         """Try to failover the current master."""
-        if departing_unit_name != self.current_master:
+        if self._k8s_hostname(departing_unit_name) != self.current_master:
             # No failover needed
             return
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -97,9 +97,14 @@ class RedisK8sCharm(CharmBase):
     def _leader_elected(self, event) -> None:
         """Handle the leader_elected event.
 
-        If no password exists, a new one will be created for accessing Redis. This password
-        will be stored on the peer relation databag.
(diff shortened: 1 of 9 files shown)


tests/integration/test_charm.py
```diff
@@ -293,14 +295,87 @@ async def test_scale_up_replication_after_failover(ops_test: OpsTest):
...
+    # INITIAL SETUP #
+    # Sanity check that the added unit on the previous test is not a master
+    assert last_redis.execute_command("ROLE")[0] != "master"
+
+    # Make the added unit a priority during failover
+    last_redis.execute_command("CONFIG SET replica-priority 1")
+    # Failover so the last unit becomes master
+    sentinel.execute_command(f"SENTINEL FAILOVER {APP_NAME}")
+    # Give time so sentinel updates information of failover
+    time.sleep(3)
```
WRFitch, line 327: Is this necessary, given that the next command is another wait?

WRFitch (approved):
LGTM, one small note about waits in testing.

tests/integration/test_charm.py
zmraul: The issue is that querying info (`"SENTINEL MASTER redis-k8s"`) right after issuing the `failover` command can result on info not being updated yet. Meaning that the next block passes immediately, and the `role` assert might fail.

src/charm.py
```diff
@@ -102,6 +105,19 @@ def _leader_elected(self, _) -> None:
                 "Initial replication, setting leader-host to {}".format(self.unit_pod_hostname)
```
MiaAltieri, line 117: I might have missed the PR where current_master was added, but I think tracking the master is a controversial practice 🤷 but idc

src/charm.py
```diff
@@ -102,6 +105,19 @@ def _leader_elected(self, _) -> None:
...
             )
             self._peers.data[self.app][LEADER_HOST_KEY] = self.unit_pod_hostname
+        else:
```
MiaAltieri, line 120: nit: (do this if you want or ignore) 
code might look nicer if you do:
```
        if not self._get_password():
            logger.info("Creating password for application")
            self._peers.data[self.app][PEER_PASSWORD_KEY] = self._generate_password()
        if not self.get_sentinel_password():
            logger.info("Creating sentinel password")
            self._peers.data[self.app][SENTINEL_PASSWORD_KEY] = self._generate_password()
        if self.current_master is None:
            logger
            self._peers
             return 
        
        # code for a new leader being elected 
        # TODO extract to method shared with relation_departed
            self._update_application_master()
            self._update_quorum()
            try:
                self._is_failover_finished()
            except (RedisFailoverCheckError, RedisFailoverInProgressError):
                logger.info("Failover didn't finish, deferring")
                event.defer()
                return

            logger.info("Resetting sentinel")
            self._reset_sentinel()
```

src/charm.py
```diff
@@ -477,13 +531,59 @@ def _update_application_master(self) -> None:
...
+        """Try to failover the current master."""
+        if self._k8s_hostname(departing_unit_name) != self.current_master:
+            # No failover needed
+            return
+
+        with self.sentinel.sentinel_client() as sentinel:
+            try:
+                sentinel.execute_command(f"SENTINEL FAILOVER {self._name}")
+            except RedisError:
+                raise
```
MiaAltieri: since you're not doing anything with RedisError I am pretty sure you can just do 
```suggestion
sentinel.execute_command(f"SENTINEL FAILOVER {self._name}")
```

src/sentinel.py
```diff
@@ -122,11 +128,19 @@ def _copy_file(self, path: str, rendered: str, container: str) -> None:
             logger.warning("Can't connect to {} container".format(container))
             return
 
+        if not container.exists(CONFIG_DIR):
```
MiaAltieri, line 133: maybe include a comment about the bug to update this code later? but idk I dont see much wrong with having a mkdir and then a push (instead of just a push (once they fix the bug))

MiaAltieri (changes requested):

MiaAltieri (commented):
looking good!

src/charm.py
paulomach: Indeed. To be on the safe side, the best would be update the master right before the test, if feasible.

src/charm.py
```diff
@@ -161,6 +181,43 @@ def _peer_relation_handler(self, event):
         # Update quorum for all sentinels
         self._update_quorum()
 
+        self.unit.status = ActiveStatus()
+
+    def _peer_relation_departed(self, event):
+        """Handle relation for leaving units."""
+        if not self.unit.is_leader():
```
paulomach, line 192: what would happen when the leader is the departing unit?

src/charm.py
```diff
@@ -477,13 +531,59 @@ def _update_application_master(self) -> None:
         """Use Sentinel to update the current master hostname."""
         info = self.sentinel.get_master_info()
         logger.debug(f"Master info: {info}")
+        if info is None:
```
paulomach, line 538: nit:

```suggestion
        if not info:
```

paulomach (approved):
Just some nits

src/charm.py
```diff
@@ -477,13 +531,59 @@ def _update_application_master(self) -> None:
...
+            logger.warning("Could not update current master")
+            return
 
         self._peers.data[self.app][LEADER_HOST_KEY] = info["ip"]
 
+    def _sentinel_failover(self, departing_unit_name: str) -> None:
+        """Try to failover the current master."""
```
marcoppenheimer: I would add to the docstring that this is a 'leader-only' function.

src/charm.py
```diff
@@ -492,11 +592,9 @@ def _update_quorum(self) -> None:
...
                 try:
-                    sentinel.execute_command(
-                        f"SENTINEL SET {self._name} quorum {self.sentinel.expected_quorum}"
-                    )
-                    logger.info(f"Quorum set to {self.sentinel.expected_quorum} on {hostname}")
-                except ConnectionError as e:
+                    logger.debug("Sending {} to sentinel at {}".format(command, hostname))
+                    sentinel.execute_command(command)
+                except (ConnectionError, TimeoutError) as e:
                     logger.error("Error connecting to instance: {} - {}".format(hostname, e))
```
marcoppenheimer, line 603: What happens to cause these exceptions to be raised? Do we expect this method call to fail silently?

tests/integration/test_charm.py
```diff
@@ -293,14 +295,87 @@ async def test_scale_up_replication_after_failover(ops_test: OpsTest):
         assert client.get("testKey") == b"myValue"
         client.close()
 
-    # TODO: scale-down to reset the model once scale-down logic is added
+
+@pytest.mark.run(after="test_scale_up_replication_after_failover")
+@pytest.mark.scaling
+async def test_scale_down_departing_master(ops_test: OpsTest):
```
marcoppenheimer, line 301: Possibly related to my other question. 
Is there value in having a test where we scale down the current Juju leader, which is also the current quorum leader?

tests/unit/test_charm.py
```diff
@@ -332,3 +354,63 @@ def test_non_leader_unit_as_replica(self):
...
+    def test_application_data_update_after_failover(self, execute_command):
+        # Custom responses to Redis `execute_command` call
+        def my_side_effect(value: str):
+            if value == "ROLE":
+                return ["non-master"]
+            if value == f"SENTINEL CKQUORUM {self.harness.charm._name}":
+                return "OK"
+            if value == f"SENTINEL MASTER {self.harness.charm._name}":
+                return ["ip", "different-leader"]
+
```
marcoppenheimer: Massively unhelpful suggestion. But for these sorts of mappings, I find it readable to do:


```suggestion
mapping = {
    "ROLE": ["non-master"],
    f"SENTINEL CKQUORUM {self.harness.charm._name}": "OK",
    f"SENTINEL MASTER {self.harness.charm._name}": ["ip", "different-leader"],
}

return mapping.get(str, "??")

```

src/charm.py
zmraul: `current_master` is used here as a way to know that the application is being deployed for the first time. Once is set, relation events will query sentinel to check if the tracked master is correct, and update if they don't match.
I will update the comments though, to make it more clear.

src/charm.py
zmraul: Yeah I was on the fence about making it explicit or not. I'll remove it since `RedisError` is being handled on the calling method.

src/charm.py
zmraul: I think I'll leave it as `if x is None` because that's an explicit return from `get_master_info()` when something wrong happens.

src/charm.py
zmraul: The problem is that I need to broadcast commands to all sentinels. But that includes sentinels that are gone on scaling down operations (when querying units in the relation you get departing units as well).

src/charm.py
zmraul: formatting might be messed up? 

edit: nvm I'm blind :)

src/charm.py
zmraul: So, chances are that sentinel will do a failover on its own, since juju leader will take some time to execute. If juju leader reelection is super fast, and this event is triggered before units depart, `sentinel reset` will do nothing, as departing units are still there. The quorum will be correctly updated, which is a more critical part of the failover process. It's not a big deal, specially in k8s. It only means that they store a pointer to a hostname that is no longer there. 

If units have not departed yet though, the new juju leader will receive the `relation_departed` event.

I could do a check on master info, to see if the flag `o_down` (objective down, meaning that sentinels have reached quorum and agree that the master is down) is there and force a failover immediately. But that flag means that a failover is going to happen pretty soon anyways.

tests/integration/test_charm.py
zmraul: I would love to add this. Is there an example somewhere on how to test this condition consistently?

tests/unit/test_charm.py
paulomach: I like that!

shayancanonical (approved):

src/exceptions.py
```diff
@@ -0,0 +1,31 @@
...
+        return "<{}.{}>".format(type(self).__module__, type(self).__name__)
+
+    @property
+    def message(self):
+        """Return the message passed as an argument."""
+        return self.args[0]
+
+
+class RedisFailoverInProgressError(RedisOperatorError):
+    """Exception raised when creating a user fails."""
```
marceloneppel: Does it really happen when creating a user fails or the description should be changed?

marceloneppel (approved):

src/exceptions.py
zmraul: Maybe I copied the exception layout from the Mysql charm, and maybe I forgot to update this description. :eyes:

PR #43, opened 2022-08-10
Fix juju version to avoid action errors

zmraul:


MiaAltieri (approved):
best thing I have ever seen ✅

paulomach (approved):
incredible

PR #44, opened 2022-08-11
Pin oci-image revision to release

zmraul:
Uses a specific charmhub resource for the redis oci image.


paulomach (approved):
amazing

marceloneppel (approved):
LGTM! I should do the same in PostgreSQL repos.

PR #47, opened 2022-08-30
Update stored state on redis relation joined

arturo-seijas:
In the event of scaling up a consumer application, the redis library does not populate the stored state when the relation for the new unit is created. This PR addresses this issue.

This change has been tested manually deploying redis along with the indico charm.


delgod (approved):

mthaddon (changes requested):
I think this will need LIBPATCH to be incremented so that the change can be published.

arturo-seijas:
> I think this will need LIBPATCH to be incremented so that the change can be published.

Indeed. Thx, Tom

mthaddon (approved):
LGTM, thx

delgod (approved):

zmraul (approved):

Issue #45, opened 2022-08-16
Update oci image to ROCK


zmraul:
Upstream oci image should point to ROCK (ubuntu/redis:6.2) once 6.2 version is created.

PR #49, opened 2022-09-14
Update OCI image to redis 7.0

zmraul:
The charm will now use [Redis 7.0](https://hub.docker.com/r/dataplatformoci/redis/tags). Solves https://github.com/canonical/redis-k8s-operator/issues/45

The change adding `--protected-mode no` is done so Redis doesn't warn of a unsecure mode being used (the password is not used for legacy relation)

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -44,7 +44,7 @@ resources:
   redis-image:
     type: oci-image
     description: ubuntu lts docker image for redis
-    upstream: redis:6.2
+    upstream: dataplatformoci/redis:7.0-22.04_edge
   cert-file:
     type: file
     filename: redis.crt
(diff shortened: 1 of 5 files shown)


metadata.yaml
```diff
@@ -44,7 +44,7 @@ resources:
   redis-image:
     type: oci-image
     description: ubuntu lts docker image for redis
-    upstream: redis:6.2
+    upstream: dataplatformoci/redis:7.0-22.04_edge
```
marcoppenheimer, line 47: Not a blocker. How is this distinct from standard `ubuntu/redis`?

marcoppenheimer (approved):

WRFitch (approved):

metadata.yaml
zmraul: Sort of like a staging area until those images are promoted to `ubuntu/<data product>`

Issue #50, opened 2022-09-21
Update connection data on scaling events.


zmraul:
If some application is related on the `redis` interface and the redis charm has events that change the IP address of the connecting pod (a pod churn), relation data should be updated.

zmraul:
Added pod deletion tests on #57

PR #55, opened 2022-10-13
Update relation ip on primary change

zmraul:
## Issue
This PR addresses Jira ticket [DPE-620](https://warthogs.atlassian.net/browse/DPE-620), and Issue #50 

Scaling events and pod reschedules now will check if they need to update the leader information on related apps.

## Solution
Redis charm can be related to an application. If a failover happens, or the pod gets rescheduled (IP change) the related application should get updated information. To do so, extra logic has been added to handle scaling and pod deletion situations.

If a pod goes down, Juju will issue a `upgrade_charm` sequence of events. There are two situations:
- **Pod is Juju leader**: In this case, the pod can update the application databag directly. This will create a `peer_relation_changed` for the rest of the units. At that moment, all will update information on the redis relation.
- **Pod is not a Juju leader**: In this case, the unit will add a key to its own unit `peer` databag. After this, the rest of the units will receive `peer_relation_changed`, and the leader unit will be tasked with updating the application databag.

When a pod goes down, and `upgrade_charm` executes, pebble_ready for sentinel has not happened yet. For this reason it is necessary to connect to a different sentinel instance to check for failover status.

## Release Notes
- `src/charm.py`. `upgrade_charm` will now check for failovers and update accordingly. `peer_relation_changed` now makes sure a failover is not in progress before updating information.
    - `check_master`: now sentinel is used to know if master is down. The host can also be used to ask a different sentinel instance about the cluster status.

## Testing
- Updated unit tests and added an extra assertion to `test_application_data_update_after_failover`
- Integration tests will be split and updated on a following PR


src/charm.py
```diff
@@ -86,14 +87,38 @@ def _redis_pebble_ready(self, event) -> None:
             event.defer()
             return
 
-    def _upgrade_charm(self, _) -> None:
+    def _upgrade_charm(self, event) -> None:
```
marcoppenheimer: Type hint here, useful.

zmraul:
Integration tests seem to be failing because timing issues.

src/charm.py
```diff
@@ -175,6 +208,12 @@ def _peer_relation_changed(self, event):
         if self._peers.data[self.app].get("enable-password", "true") == "false":
             self._update_layer()
 
+        relation = self.model.get_relation(relation_name=REDIS_REL_NAME)
+        if relation:
+            relation.data[self.model.unit]["hostname"] = socket.gethostbyname(self.current_master)
+            if self._peers.data[self.unit].get("upgrading", "false") == "true":
+                self._peers.data[self.unit]["upgrading"] = ""
```
marcoppenheimer, line 214: If I understand correctly, you're setting a flag on all non-leader units during `upgrade-charm` for `upgrading=true`, and then you unset it on `peer_relation_changed`.

What I couldn't see is where the flag gets used for anything?
Or is setting it in and of itself supposed to trigger a `peer_relation_changed` to everyone?

If so, a single-unit deployment wouldn't get this event, is that right?

src/charm.py
```diff
@@ -565,15 +605,19 @@ def _sentinel_failover(self, departing_unit_name: str) -> None:
         reraise=True,
         before=before_log(logger, logging.DEBUG),
     )
-    def _is_failover_finished(self) -> None:
-        """Check if failover is still in progress."""
+    def _is_failover_finished(self, host="localhost") -> None:
```
delgod: Why not return True/False? why exception needed?

src/charm.py
```diff
@@ -86,14 +87,38 @@ def _redis_pebble_ready(self, event) -> None:
             event.defer()
             return
 
-    def _upgrade_charm(self, _) -> None:
+    def _upgrade_charm(self, event) -> None:
         """Handle the upgrade_charm event.
 
         Tries to store the certificates on the redis container, as new `juju attach-resource`
         will trigger this event.
```
MiaAltieri, line 95: update docstring

src/charm.py
```diff
@@ -86,14 +87,38 @@ def _redis_pebble_ready(self, event) -> None:
...
         """
         self._store_certificates()
 
+        # NOTE: This is the case of a single unit deployment. If that's the case, the charm
+        # doesn't need to check for failovers or figure out who the master is.
+        if not self._peers.units:
+            return
+
+        # Pick a different unit to connect to sentinel
+        k8s_host = self._k8s_hostname(name=list(self._peers.units)[0].name)
```
MiaAltieri, line 105: I don't know enough about k8s or redis, but does it matter which unit we connect to, what if that other unit is also down?

src/charm.py
```diff
@@ -86,14 +87,38 @@ def _redis_pebble_ready(self, event) -> None:
...
+
+        if self.unit.is_leader():
+            info = self.sentinel.get_master_info(host=k8s_host)
+            logger.debug(f"Master info: {info}")
+            logger.info(f"Unit {self.unit.name} updating master info to {info['ip']}")
+            self._peers.data[self.app][LEADER_HOST_KEY] = info["ip"]
+        else:
+            relation = self.model.get_relation(relation_name=REDIS_REL_NAME)
+            if relation:
+                self._peers.data[self.unit]["upgrading"] = "true"
```
MiaAltieri, line 123: comments explaining what explained in the PR message would be awesome here

src/charm.py
```diff
@@ -522,17 +561,17 @@ def _redis_client(self, hostname="localhost") -> Redis:
...
 
-    def _check_master(self) -> bool:
-        """Connect to the current stored master and query role."""
-        with self._redis_client(hostname=self.current_master) as redis:
-            try:
-                result = redis.execute_command("ROLE")
-            except (ConnectionError, TimeoutError) as e:
-                logger.warning("Error trying to check master: {}".format(e))
-                return False
+    def _check_master(self, host="0.0.0.0") -> bool:
```
MiaAltieri:
```suggestion
    def _master_up_to_date(self, host="0.0.0.0") -> bool:
```

MiaAltieri (commented):
Super clever way of handling, this! 🤩

Looks good, also would be nice to change the language from master to primary (feel free to do this in a different PR).

See a conversation on this here: https://chat.canonical.com/canonical/pl/kb7k7pnhbifyb8mqoubg8zb1ky

src/charm.py
zmraul: The flag is only used to trigger peer_relation_changed. Single unit deployment doesn't really need to do any of this, since you can only have one unit being the leader.

src/charm.py
zmraul: Yeah, I think that will be an edge case to this solution. But it will happen either way. We could potentially have all pods down (full cluster crash).
I guess I could change this to connect to all other units and keep the first sentinel connection that has quorum.

src/charm.py
zmraul: Initially, checking for failover was meant to be the not expected path. But It has changed to be a relevant part of how the charm checks redis health status. I can change it to be True/False.

zmraul:
> Looks good, also would be nice to change the language from master to primary (feel free to do this in a different PR).

Yeah, this will be changed, thanks for the remainder. Redis will update terminology in the future to change this as well. But at the moment some sentinel/redis commands still refer to the _master_ replica.

marcoppenheimer (approved):

MiaAltieri (approved):

PR #56, opened 2022-10-18
chore: Split integration tests

zmraul:
Split integration tests into own files and change CI/CD pipelines.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -75,15 +75,20 @@ deps =
     pytest-order
     -r{toxinidir}/requirements.txt
 commands =
-    # Generate TLS certificates to use during tests
-    sh -c "./tests/gen-test-certs.sh"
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -75,20 +75,45 @@ deps =
...
+description = Run password integration tests
+deps =
+    pytest
+    juju
+    pytest-operator
+    -r{toxinidir}/requirements.txt
+commands =
+    pytest {[vars]tst_path}/integration/test_password.py -vv --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --durations=0
+
+[testenv:integration-relation]
```
MiaAltieri, line 111: renaming to `relation-integration` will save you keystrokes, but idc what you do here

MiaAltieri (approved):

tox.ini
zmraul: Hm, I can change it to `integration-redis-relation` so it's more specific (the only interface is called `redis`)

WRFitch (approved):

PR #57, opened 2022-10-19
feat: Add pod deletion tests

zmraul:
Add pod deletion integration tests.

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -130,7 +130,7 @@ async def test_delete_non_primary_pod(ops_test: OpsTest):
 
     # Wait for `upgrade_charm` sequence
     await ops_test.model.wait_for_idle(
-        apps=[APP_NAME], status="active", timeout=1000, wait_for_exact_units=3, idle_period=120
+        apps=[APP_NAME], status="active", timeout=1000, wait_for_exact_units=3, idle_period=140
     )
 
     pod_num = get_unit_number(non_leader)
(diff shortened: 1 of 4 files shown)


tests/integration/test_charm.py
```diff
@@ -118,8 +118,62 @@ async def test_sentinels_expected(ops_test: OpsTest):
     assert sentinels_connected == NUM_UNITS
 
 
+async def test_delete_non_primary_pod(ops_test: OpsTest):
+    """Delete a pod that is not the Redis primary. Check that the deployment is healthy."""
```
marcoppenheimer: Worth adding to the docstring that you're expecting the unit to recover because it's in a k8s statefulset, and you're checking it's recovered fine and rejoined

tests/integration/test_charm.py
```diff
@@ -118,8 +118,62 @@ async def test_sentinels_expected(ops_test: OpsTest):
...
+    leader = unit_map["leader"]
+
+    pod_num = get_unit_number(leader)
+    pod_address = await get_address(ops_test, unit_num=pod_num)
+    password = await get_sentinel_password(ops_test, pod_num)
+
+    sentinel = Redis(pod_address, password=password, port=26379)
+
+    assert len(sentinel.sentinel_sentinels(service_name=APP_NAME)) == NUM_UNITS - 1
+    assert len(sentinel.sentinel_slaves(service_name=APP_NAME)) == NUM_UNITS - 1
```
marcoppenheimer, line 181: I probably misunderstood. Are you doing a `num-sentinels=num-units` deployment?

If so, after removing Redis master pod, it recovers, there's a leader election, and it would rejoin with N-units sentinels still? 

I see you're checking for N-1 slaves, which makes sense.

marcoppenheimer (approved):
I like it, left a couple questions but nice and clean.

tests/integration/test_charm.py
zmraul: In this case `sentinel_sentinels()` will give information about sentinels other than the own instance.
I'm calling it from the pod that was killed. If the new sentinel instance has discovered the other two, and it agrees on the expected number of replicas, it joined the deployment successfully after the failure. 

I might change this so the check is more explicit.

WRFitch (approved):

PR #59, opened 2022-11-23
Update relation data when the charm pod restarted in a single unit deployment

weiiwang01:
## Issue

When the Redis charm pod was restarted by Kubernetes in a single unit deployment. The relation data will not update after the pod IP is changed. This is discovered during yesterday's indico incident.

## Solution
Update the relation data when the charm detected that it's in a single unit deployment in the `upgrade-charm` event handler.

Also, fix some potential problems which may occur when there is more than one redis relation attached to a single redis charm application.

## Testing
Add some integration tests for this situation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,6 +61,24 @@ jobs:
       - name: Run integration tests
         run: tox -e integration-relation
 
+  integration-test-relation-single-unit:
+    name: Integration tests (relation, num_units=1)
+    needs:
+      - lint
+      - unit-test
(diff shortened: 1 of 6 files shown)


zmraul:
This is a really good find, the previous PR on this issue only addressed multi-unit deployments. I only did a pass at the tests, but I think having the number of units as a fixture is a clever approach. :)

jdkandersson (approved):

zmraul (approved):

deusebio (approved):
LGTM!

PR #60, opened 2022-11-30
Bump action versions and simplify release.yaml

zmraul:
Simplify `release.yaml` and bump action versions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


welpaolo (approved):
Looks good to me

deusebio (approved):
Looks good! I would probably factor out all tests into a single command out of the ci/cd pipeline such that it can be equally called on ci/cd pipeline or locally, and keep CI/CD business logic minimal. 

But this is well beyond the scope of this PR and to be discussed before hand. 

I have seen that we have a failing test in CI. Why is that? Should one open an issue on github?

zmraul:
> I have seen that we have a failing test in CI. Why is that? Should one open an issue on github?

I believe this is a timing issue where sometimes the application is not up yet. I could look into it more closely but it is a bit hard to debug, as this only happens on CI.

Issue #51, opened 2022-09-22
Add COS integration


mthaddon:
When running this charm in a production setting it would be useful to be able to integrate it with [COS](https://charmhub.io/topics/canonical-observability-stack) to get metrics/dashboards for the application.

reneradoi:
Resolved with https://github.com/canonical/redis-k8s-operator/pull/61.

As of revision 29 in `latest/edge`, this charm now supports COS lite. Please make sure to pull the new image when updating, e.g. with `juju refresh ... --resource redis-image=ghcr.io/canonical/charmed-redis:7.2.5-22.04-edge`.

Redis metrics can be found in the dashboard `Redis Dashboard for Prometheus Redis Exporter` under `General`.

Other than that, the following changes are included in this revision:
- update to version 7.2.5 of redis
- redis now runs in `appendonly` mode by default in order to avoid data loss (see also: https://redis-stack.io/docs/management/persistence/)

PR #61, opened 2022-11-30
COS integration

zmraul:
Addresses [DPE-1021](https://warthogs.atlassian.net/browse/DPE-1021)

This PR shows how a DB specific exporter and some default dashboards can be added to a charm. Addresses #51 

Since this is a k8s charm, the exporter has been added as an extra container to the deployment. Ideally, exporters will come packed with the workload container (specifically node and process exporter).

[DPE-1021]: https://warthogs.atlassian.net/browse/DPE-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reneradoi (dismissed):

reneradoi:
This PR is now ready for review.

metadata.yaml
```diff
@@ -39,12 +47,18 @@ containers:
...
 
 resources:
   redis-image:
     type: oci-image
     description: ubuntu lts docker image for redis
     upstream: ghcr.io/canonical/charmed-redis:7.0-22.04_edge
+  exporter:
+    type: oci-image
+    description: redis exporter for prometheus metrics
+    upstream: docker.io/oliver006/redis_exporter:v1.59.0
```
delgod: Yes, it is the right thing in most cases, but I believe we should control all the artefacts in Canonical solutions.
Could we please add the exporter into your image, without using 3rd party image?
(in such a case, the additional resource and side-car(layer) are not needed).

delgod (changes requested):

src/charm.py
```diff
@@ -52,6 +58,23 @@ def __init__(self, *args):
...
         self.redis_provides = RedisProvides(self, port=REDIS_PORT)
         self.sentinel = Sentinel(self)
+        self.exporter = Exporter(self)
+        self.metrics_endpoint = MetricsEndpointProvider(
+            self,
+            jobs=[
+                {
+                    "static_configs": [
+                        {
+                            "targets": ["*:9121"],
```
MiaAltieri: nit constant

src/charm.py
```diff
@@ -302,6 +326,15 @@ def _update_layer(self) -> None:
             self.unit.status = WaitingStatus("Waiting for Pebble in workload container")
             return
 
+        if not container.exists(LOG_DIR):
+            container.make_dir(
+                LOG_DIR,
+                make_parents=True,
+                permissions=0o770,
+                user="redis",
+                group="redis",
```
MiaAltieri: nit const for "redis" too (if its too annoying feel free to ignore)

src/redis_exporter.py
```diff
@@ -0,0 +1,75 @@
...
+            "services": {
+                "exporter": {
+                    "override": "replace",
+                    "summary": "Redis exporter service",
+                    "command": "/redis_exporter",
+                    # "user": "redis",
+                    # "group": "redis",
+                    "startup": "enabled",
+                    "environment": {
+                        "REDIS_PASSWORD": self.charm._get_password(),
```
MiaAltieri: since this relies on the charm password, we must be sure that we make a call to `_update_exporter_layer ` on password change events (if the charm handles them)

MiaAltieri (changes requested):
Looks great! +1 to Mykolas earlier point

And if possible can you add a test suite similar to [this one](https://github.com/canonical/mongodb-k8s-operator/blob/6/edge/tests/integration/metrics_tests/test_metrics.py)? A separate PR would also be fine

src/charm.py
reneradoi: done

src/charm.py
reneradoi: done

metadata.yaml
reneradoi: This charm now uses the [charmed-redis-rock](https://github.com/canonical/charmed-redis-rock/blob/main/rockcraft.yaml), which includes the `redis_exporter` binary based on [this fork](https://github.com/canonical/redis_exporter/tree/v1.60.0) of the original code.

src/redis_exporter.py
reneradoi: This charm does not handle passwort changes, so no need for action here.

src/charm.py
```diff
@@ -29,6 +29,7 @@
     LEADER_HOST_KEY,
     LOG_DIR,
     LOG_FILE,
+    METRICS_PORT,
     PEER,
```
MiaAltieri, line 33: Note for future Redis developers, this will grow and slowly become huge, maybe in the future it will be worth it to have a constant's "class" that way you can do things like `Constants.METRICS_PORT ` instead of importanting all of the constants.

Obv not a priority at all, but maybe a nice task for a future onboarder or a quick clean up task

metadata.yaml
```diff
@@ -47,18 +47,12 @@ containers:
...
 resources:
   redis-image:
     type: oci-image
     description: ubuntu lts docker image for redis
-    upstream: ghcr.io/canonical/charmed-redis:7.0-22.04_edge
-  exporter:
-    type: oci-image
-    description: redis exporter for prometheus metrics
-    upstream: docker.io/oliver006/redis_exporter:v1.59.0
+    upstream: ghcr.io/canonical/charmed-redis:7.2.5-22.04-edge
```
MiaAltieri, line 55: 🥳 nice work

MiaAltieri (commented):
Looks great! Do you plan to add integrations tests in this PR or in a future one?

delgod (approved):

reneradoi:
> And if possible can you add a test suite similar to [this one](https://github.com/canonical/mongodb-k8s-operator/blob/6/edge/tests/integration/metrics_tests/test_metrics.py)? A separate PR would also be fine

I've added some test coverage for the metrics exporter here: https://github.com/canonical/redis-k8s-operator/blob/39b9098effd5194320d039486f39163d07334d3c/tests/integration/test_charm.py#L86-L98

MiaAltieri (approved):
Thank you Rene!

DEVELOPMENT.md
```diff
@@ -36,7 +36,7 @@ to discover the IP Redis is running behind. The output will have lines like:
 
 To retrieve the password to access the database, use the `get-initial-admin-password` action:
 
-    juju run-action redis-k8s/0 get-initial-admin-password --wait
+    juju run redis-k8s/0 get-initial-admin-password --wait
```
Mehdi-Bendriss: we unfortunately need to rollback this change (and the ones after it), because we currently offer the Redis charm fir juju `2.9.x` and `juju run-action`  only got renamed in 3.x

Mehdi-Bendriss (commented):
Thanks René, I have a minor comment on the README

DEVELOPMENT.md
reneradoi: Oh ok, I didn't know that, just changed this on the fly because my juju client complained when I pasted the commands.

reneradoi: done

Mehdi-Bendriss (approved):

PR #63, opened 2023-02-17
Cache charmcraft pack container, skip unstable tests except on schedule

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/146

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
 
   build:
     name: Build charms
-    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v1
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v2
 
   integration-test:
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,7 +1,13 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
 name: Tests
+
 on:
-  workflow_call:
   pull_request:
+  schedule:
+    - cron: '53 0 * * *' # Daily at 00:53 UTC
+  # Triggered on push to branch "main" by .github/workflows/release.yaml
+  workflow_call:
 
 jobs:
   lint:
@@ -10,10 +16,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Install dependencies
+      - name: Install tox
+        # TODO: Consider replacing with custom image on self-hosted runner OR pinning version
         run: python3 -m pip install tox
       - name: Run linters
-        run: tox -e lint
+        run: tox run -e lint
 
   unit-test:
     name: Unit tests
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,7 +1,13 @@
+# Copyright 2022 Canonical Ltd.
```
zmraul, line 1:
```suggestion
# Copyright 2023 Canonical Ltd.
```

tests/integration/conftest.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
```
zmraul, line 2:
```suggestion
# Copyright 2023 Canonical Ltd.
```

tox.ini
```diff
@@ -2,27 +2,27 @@
 # See LICENSE file for licensing details.
 
 [tox]
```
zmraul, line 4: Is it needed to ask for tox v4 minimum here?

zmraul (approved):
LGTM. Some minor comments

.github/workflows/ci.yaml
carlcsaposs-canonical: Some of the code was written in 2022

tox.ini
carlcsaposs-canonical: tox v4 or later is required. Should I add `min_version` to tox.ini? One downside of that is that if tox v4 is not installed, it will install tox inside tox, which could make debugging tox/virtual env issues harder

zmraul: Hm, yeah in that case it's better to leave it as it is. But I definitely got into a small rabbit hole when trying out the new tox changes with v3.

carlcsaposs-canonical:
Added changes from https://github.com/canonical/mysql-operator/pull/134

DEVELOPMENT.md
```diff
@@ -102,11 +102,11 @@ To see the logs for a specific pod:
 The Python operator framework includes a very nice harness for testing
 operator behaviour without full deployment.
 
-    tox -e fmt           # update your code according to linting rules
-    tox -e lint          # code style
-    tox -e unit          # unit tests
-    tox -e integration   # integration tests
-    tox                  # runs 'lint' and 'unit' environments
+    tox run -e fmt           # update your code according to linting rules
```
deusebio: should this be `format`?

deusebio (approved):
LGTM

@zmraul is it normal that the CI is failing? Should we have an issue/ticket to fix this?

DEVELOPMENT.md
carlcsaposs-canonical: Fixed in ac62ad6bf65e42da4a05d27c4c2606695421c9e3

zmraul (changes requested):
@carlcsaposs-canonical @deusebio I haven't looked too much into this, but I noticed that we have two `conftest` files. There is one uner `tests/conftest.py` that handles passing a number of units to the integration tests, and the new one on `tests/integration/conftest.py`.
It's a guess though, I only took a quick glance at the errors on CI and the only significant things is that `build_and_deploy` tests fail.

carlcsaposs-canonical:
> @carlcsaposs-canonical @deusebio I haven't looked too much into this, but I noticed that we have two `conftest` files. There is one uner `tests/conftest.py` that handles passing a number of units to the integration tests, and the new one on `tests/integration/conftest.py`. It's a guess though, I only took a quick glance at the errors on CI and the only significant things is that `build_and_deploy` tests fail.

Having multiple `conftest.py` files is standard practice. From a quick glance, I don't see any issues with the `conftest` files

zmraul:
These changes have been addressed on #80. Thanks!

PR #67, opened 2023-06-08
[DPE-2097] Trigger relation changed when pod get restarted

arturo-seijas:
In the event of a pod restart on the same node the upgrade event is not fired. The IP might change, so the data needs to be propagated

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -89,7 +89,7 @@ class RedisK8sCharm(CharmBase):
         # The IP might change, so the data needs to be propagated
         for relation in self.model.relations[REDIS_REL_NAME]:
             relation.data[self.model.unit]["hostname"] = socket.gethostbyname(
-                self.unit_pod_hostname
+                self.current_master
             )
 
     def _upgrade_charm(self, event: UpgradeCharmEvent) -> None:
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -85,6 +85,12 @@ def _redis_pebble_ready(self, event) -> None:
         if not isinstance(self.unit.status, ActiveStatus):
             event.defer()
             return
+        # In the event of a pod restart on the same node the upgrade event is not fired.
+        # The IP might change, so the data needs to be propagated
+        for relation in self.model.relations[REDIS_REL_NAME]:
+            relation.data[self.model.unit]["hostname"] = socket.gethostbyname(
+                self.unit_pod_hostname
```
delgod: I think it is more correct to put leader IP
```suggestion
                self.current_master
```

delgod (changes requested):

src/charm.py
zmraul: Good catch! Yes, you need to announce the IP using `current_master`. Same as here: https://github.com/canonical/redis-k8s-operator/blob/5b796003e0e99a6ecdd4fd64b83f4c6cc1102e2e/lib/charms/redis_k8s/v0/redis.py#L103

delgod (approved):

zmraul (approved):

PR #68, opened 2023-06-09
[DPE-2097] Fix integration tests

zmraul:
Fix integration tests.

Postgresql channel needs to be changed once https://github.com/canonical/postgresql-operator/pull/139 is merged and released

1) Postgresql channel `latest/stable` can't be deployed on agent version `2.9.29`
2) Redis scale down only works on agent version <=`2.9.42` and integration tests need to be pinned on `2.9.29`

For the two reasons above, I've changed the agents depending on the tests. I'll add an issue to fix scale down implementation for agent > `2.9.42`

### DIFF
--- a/tests/integration/test_scaling.py
+++ b/tests/integration/test_scaling.py
@@ -146,6 +146,7 @@ async def test_scale_down_departing_master(ops_test: OpsTest):
         timeout=60,
     )
     assert last_redis.execute_command("ROLE")[0] == "master"
+    sentinel.close()
     last_redis.close()
 
     # SCALE DOWN #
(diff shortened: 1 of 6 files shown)


arturo-seijas (approved):
LGTM

tests/integration/test_scaling.py
```diff
@@ -158,6 +159,10 @@ async def test_scale_down_departing_master(ops_test: OpsTest):
         assert client.get("testKey") == b"myValue"
         client.close()
 
+    address = await get_address(ops_test, unit_num=0)
+    sentinel = Redis(address, port=26379, password=sentinel_password, decode_responses=True)
+    sentinel.execute_command(f"SENTINEL RESET {APP_NAME}")
+    time.sleep(10)
```
delgod: I think HA should work fine without these lines

delgod (changes requested):

deusebio (approved):
LGTM

tests/integration/test_scaling.py
zmraul: The issue seems to happen because of timing on sentinel side apparently, but Redis does update the primary. This is checked on the [relation tests here](https://github.com/canonical/redis-k8s-operator/pull/68/files#diff-1d7a38961d1e0695f63f1958ae253a1228160723122667bb47c158042971e30bR124), where the primary pod is deleted while being related to discourse.

delgod (approved):

PR #75, opened 2023-11-15
Support aliasing the redis relation

### LINKED ISSUE #30: Support relation aliases and multiple relations
Right now the redis client library does not support adding aliases to the relations and hence, multiple relations with the same consumer charm. The relation event name is hardcoded:

```
  class RedisRequires(Object):    
      def __init__(self, charm, _stored):

          """A class implementing the redis requires relation."""
          super().__init__(charm, "redis")
          self.framework.observe(charm.on.redis_relation_changed, self._on_relation_changed)
          self.framework.observe(charm.on.redis_relation_broken, self._on_relation_broken)
          self._stored = _stored
          self.charm = charm
```

We request this change because we've identified use cases in which two different redis clusters need to be consumed.

### LINKED ISSUE #74: Remove reliance on StoredState in relation library
The current relation library for redis relies on StoredState (per [this code](https://github.com/canonical/redis-k8s-operator/blob/main/lib/charms/redis_k8s/v0/redis.py#L63-L70) you pass it into the `RedisRequires` constructor).

The [juju documentation on StoredState](https://juju.is/docs/sdk/stored-state-uses-limitations) says "charm authors should avoid state when they can". It would be great if the library could be updated to not rely on StoredState.

arturo-seijas:
## Issue
Closes #30 and closes #74 

## Solution
Added support for aliases for the redis requirers and some method in the lib to retrieve the already existing relation data.
Tested for regression on the [indico charm](https://github.com/canonical/indico-operator)


zmraul (approved):

lib/charms/redis_k8s/v0/redis.py
```diff
@@ -89,19 +93,44 @@ def _on_relation_broken(self, event):
...
+        Returns:
+            Dict: dict containing the relation data.
+        """
+        relation = self.model.get_relation(self.relation_name)
+        if not relation or not relation.units:
+            return None
+        unit = next(iter(relation.units))
+        return relation.data[unit]
+
+    def get_url(self) -> Optional[str]:
```
deusebio: **niptick** maybe one could use property instead of methods, given there is no input

deusebio (approved):
LGTM

PR #83, opened 2024-05-15
Remove stored state from the redis library

arturo-seijas:
## Issue
#74 

## Solution
Removes stored state, breaking compatibility


reneradoi:
The update to the lib version causes the `charmcraft pack` command to fail with:
```
Library 'lib/charms/redis_k8s/v0/redis.py' metadata field LIBAPI is different from the version in the path.                                             
Failed to build charm for bases index '0'.
```
both locally and in CI.

reneradoi (approved):

Mehdi-Bendriss (approved):
I understand that we are breaking backward compatibility by keeping the major version of the lib as is, but we are purposefully doing so as we consider the use of the `storedstate` object in the relation to be a bug.

Issue #90, opened 2024-07-02
redis-server does not start on Canonical Kubernetes


vmpjdc:
Hi,

I noticed that redis-k8s does not work on Canonical Kubernetes (channel `1.30/beta`, revision 64).

```
Model           Controller                      Cloud/Region            Version  SLA          Timestamp         
stg-netbox-k8s  juju-controller-34-staging-ps6  stg-netbox-k8s/default  3.4.2    unsupported  21:13:30Z         
                                                                                                                        
App        Version  Status   Scale  Charm      Channel        Rev  Address         Exposed  Message             
redis-k8s           waiting      1  redis-k8s  latest/edge     32  10.152.183.100  no       installing agent    
                                                                                                                        
Unit          Workload  Agent  Address     Ports  Message                                                               
redis-k8s/0*  error     idle   10.1.1.63          hook failed: "storage-attached"                                       
```

Digging into the deployment, I found this:

```
$ kubectl exec -t -n stg-netbox-k8s redis-k8s-0  -c redis -- head /var/log/redis/redis-server.log
13:C 01 Jul 2024 05:15:06.233 * oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
13:C 01 Jul 2024 05:15:06.233 * Redis version=7.2.5, bits=64, commit=00000000, modified=0, pid=13, just started
13:C 01 Jul 2024 05:15:06.233 * Configuration loaded
13:M 01 Jul 2024 05:15:06.233 * Increased maximum number of open files to 10032 (it was originally set to 1024).
13:M 01 Jul 2024 05:15:06.233 * monotonic clock: POSIX clock_gettime
13:M 01 Jul 2024 05:15:06.234 * Running mode=standalone, port=6379.
13:M 01 Jul 2024 05:15:06.234 * Server initialized
13:M 01 Jul 2024 05:15:06.240 # Can't open or create append-only dir appendonlydir: Permission denied
18:C 01 Jul 2024 05:15:06.839 * Redis version=7.2.5, bits=64, commit=00000000, modified=0, pid=18, just started
$ _
```

When the Juju storage is provisioned on a Canonical Kubernetes cluster, the permissions of `/var/lib/redis` are as follows:

```
root@redis-k8s-0:/# ls -ld /var/lib/redis/
drwxr-xr-x 4 root root 4096 Jul  2 01:58 /var/lib/redis/
root@redis-k8s-0:/# _
```

[...]

PR #93, opened 2024-07-05
[DPE-4808] Initialize directory structure on startup

reneradoi:
## Issue
The fix in https://github.com/canonical/redis-k8s-operator/pull/91 for issue https://github.com/canonical/redis-k8s-operator/issues/90 did not solve the problem, as the Juju-created storage volume is obviously mounted on top of the existing directory on the image.

## Solution
To make sure all required directories have all required permissions, no matter the file system defaults of the cloud it is deployed to, we will initialize the directory structure on startup and create the directories with the required permissions within the container.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -337,6 +337,15 @@ class RedisK8sCharm(CharmBase):
                 group=REDIS_USER,
             )
 
+        if not container.exists(WORKING_DIR):
+            container.make_dir(
+                WORKING_DIR,
+                make_parents=True,
+                permissions=0o770,
+                user=REDIS_USER,
+                group=REDIS_USER,
+            )
+
         if not self._valid_app_databag():
             self.unit.status = WaitingStatus("Waiting for peer data to be updated")
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -102,7 +102,6 @@ class RedisK8sCharm(CharmBase):
 
         Updates the Pebble layer if needed.
         """
-        self._initialize_directory_structure()
         self._store_certificates()
         self._update_layer()
 
@@ -329,6 +328,8 @@ class RedisK8sCharm(CharmBase):
             self.unit.status = WaitingStatus("Waiting for Pebble in workload container")
             return
 
+        self._initialize_directory_structure()
+
         if not self._valid_app_databag():
             self.unit.status = WaitingStatus("Waiting for peer data to be updated")
             return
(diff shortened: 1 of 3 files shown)


tests/integration/test_scaling.py
```diff
@@ -51,7 +51,7 @@ async def test_build_and_deploy(ops_test: OpsTest):
     assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
 
 
-@pytest.mark.run(before="test_scale_down_departing_master")
+@pytest.mark.abort_on_fail
```
reneradoi, line 54: fyi: there were warnings in the ci runs for this, not in any relation to the actual issue.

src/charm.py
```diff
@@ -356,6 +349,33 @@ def _update_layer(self) -> None:
...
+            container.make_dir(
+                WORKING_DIR,
+                make_parents=True,
+                permissions=0o770,
+                user=REDIS_USER,
+                group=REDIS_USER,
+            )
+        else:
+            try:
+                container.exec(["chown", "-R", f"{REDIS_USER}:{REDIS_USER}", WORKING_DIR])
```
Mehdi-Bendriss, line 375: Interesting, does this mean that the user/group set in `container.make_dir`isn't applied recursively?

Mehdi-Bendriss (approved):

src/charm.py
reneradoi: No, the problem was rather that the directory was already there but ownership was changed on deployment, that means it belonged to `root` and user `redis` could not create files in there. This obviously only happens with canonical k8s (as described in the issue), as I never had any issues testing this with microk8s.

The directory is already there because it is created in the rock (where I thought was the problem initially). Maybe we could remove it from there (as this layer obviously can be overwritten, depending on the defaults of the k8s it is deployed to).

src/charm.py
```diff
@@ -356,6 +349,33 @@ def _update_layer(self) -> None:
 
         self.unit.status = ActiveStatus()
 
+    def _initialize_directory_structure(self) -> None:
+        """Make sure all required directories for redis are available on startup."""
+        container = self.unit.get_container("redis")
+
+        if not container.exists(LOG_DIR):
```
MiaAltieri, line 356: Hmm, i think this would be better inside the snap/rock. 

But I could be wrong here. What do you think?

MiaAltieri (approved):

src/charm.py
reneradoi: In general yes, but in this case the directory is mounted as storage by the k8s storage provider and too strict defaults are applied. To override these defaults with the permissions we need, the directory operations are made in the charm, not in the rock.

Issue #95, opened 2024-08-20
Add Sentinel Master units ip to the app databag


alithethird:
When multiple units of `Redis` is deployed the units put their ip into the unit databag and consumer charms can not find which unit is the leader. An example bug is here: https://github.com/canonical/discourse-k8s-operator/issues/268

This also fixes the issue where after a reboot the consumer charm doesn't refresh the redis ip. Ip refreshes in the unit databag but the charm doesn't read that because its cashed. An example bug is here: https://bugs.launchpad.net/juju/+bug/2063087

Adding the Sentinel Master units ip to the app databag would fix both of these bugs.

reneradoi:
Hello @alithethird thanks for creating this request. Just for clarification: You need the Redis Sentinel leader information, not the `juju` leader, correct?

alithethird:
Hi @reneradoi. Thanks for the answer. Sorry I am not very familiar with and thought they were the same. I need the one that has write access to the database. That is the Sentinel leader I guess?

reneradoi:
Yes @alithethird that would be the Sentinel leader.

alithethird:
Thanks for clarifying that. I will update the issue accordingly.

AmberCharitos:
This is also impacting the Commercial Systems team deployment of Superset. Commenting here for visibility. I see there is a PR to resolve so will monitor that.

reneradoi:
New version has been released as revision 35 to Charmhub.

alithethird:
Thanks a lot!

reneradoi:
Closing the issue as this has been provided with https://github.com/canonical/redis-k8s-operator/pull/96.

PR #96, opened 2024-08-21
[DPE-5200] add leader info to relation data

reneradoi:
## Issue
With https://github.com/canonical/redis-k8s-operator/issues/95 the Discourse team requests to have the Redis Sentinel leader information in the relation data.

## Solution
On updates to the Sentinel leader and when a new relation to redis is created, add the leader information to the relation data of all related applications.

## Note
It also fixes an issue with `metadata.yaml`, which caused builds to fail: With a recent `charmcraft` update, the summary in `metadata.yaml` must not be longer than 200 chars anymore.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -12,20 +12,12 @@ tags:
   - kubernetes
   - redis
 summary: >
-  Redis is an open source (BSD licensed), in-memory data structure store, used
-  as a database, cache, and message broker. Redis provides data structures
-  such as strings, hashes, lists, sets, sorted sets with range queries,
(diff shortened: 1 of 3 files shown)


alithethird:
Hi Rene! Thanks for the changes. Do you plan to also update the library to get the master ip directly? I think also the url function in the library should be updated too.

Mehdi-Bendriss (approved):
In general, we will rewrite the external relations lib / contract to align with the [DP interfaces](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces) - as those are legacy interfaces, so I see no harm in this PR for the time being.

skourta (approved):

Issue #70, opened 2023-06-14
Scale down implementation does not work on juju agent >= `2.9.43`


zmraul:
With the release of this fix for juju: https://bugs.launchpad.net/juju/+bug/1951415 scaling down operations now don't immediately terminate the pod.

The current implementation of scale down on redis uses the instant failover (in case of primary scale down) that happened before the fix. 

If redis is deployed on agent version `2.9.43`, and a scale down operation is executed on the primary, redis server will elect a new primary, but sentinel doesn't get a proper `SENTINEL RESET` to relearn the status of the deployment.

PR #99, opened 2024-09-24
Adjust sentinel config directory permission to proactively prepare for pebble v1.15

shayancanonical:
## Issue
Juju 3.6 will upgrade from pebble v1.10.2 to v1.15.0 which includes the following PR: https://github.com/canonical/pebble/pull/423

TLDR; pebble did not account for umask previous to this PR. so if one uses `pebble mkdir` with permissions `770` or `777`, umask will de-escalate privileges to `755`. however, with this PR, the permissions for a newly created directory will be exactly what one specifies

## Solution
Since sentinel config directory (created with `770`) is actually being de-escalated to `750`, change the permissions to it explicitly to avoid surprises when upgrading to juju 3.6

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,7 +68,7 @@ jobs:
           provider: microk8s
           channel: "1.28-strict/stable"
           juju-channel: 2.9/stable
-          bootstrap-options: "--agent-version 2.9.50"
+          bootstrap-options: "--agent-version 2.9.49"
       - name: Download packed charm(s)
         uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -146,7 +146,7 @@ jobs:
           provider: microk8s
           channel: "1.28-strict/stable"
           juju-channel: 2.9/stable
-          bootstrap-options: "--agent-version 2.9.29"
+          bootstrap-options: "--agent-version 2.9.50"
```
zmraul: Beware scaling tests won't work on this version because of https://github.com/canonical/redis-k8s-operator/issues/70

reneradoi (approved):

.github/workflows/ci.yaml
reneradoi: Have reverted this.

Mehdi-Bendriss (approved):
Thanks Shayan!

PR #100, opened 2024-10-14
chore(lib): Updated url with the leader host

alithethird:
## Issue

Since [DPE-5200] add leader info to relation data [PR](https://github.com/canonical/redis-k8s-operator/pull/96) we need to update a lot of our(IS DevOps) charms to use the `leader-host`. We also use the `.url` property but since it uses the unit ip we have to construct it in the charm. 
## Solution
Constructing the `.url` property in the library instead of changing multiple number of charms.

[DPE-5200]: https://warthogs.atlassian.net/browse/DPE-5200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/redis_k8s/v0/redis.py
+++ b/lib/charms/redis_k8s/v0/redis.py
@@ -39,7 +39,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version.
-LIBPATCH = 6
+LIBPATCH = 7
 
 logger = logging.getLogger(__name__)
 
@@ -102,6 +102,12 @@ class RedisRequires(Object):
         if not relation_data:
             return None
         redis_host = relation_data.get("hostname")
+        relation = self.model.get_relation(self.relation_name)
+        relation_app_data = relation.data.get(relation.app)
+        try:
+            redis_host = relation_app_data.get("leader-host", redis_host)
+        except KeyError:
+            pass
         redis_port = relation_data.get("port")
         return f"redis://{redis_host}:{redis_port}"
 
--- a/lib/charms/redis_k8s/v0/redis.py
+++ b/lib/charms/redis_k8s/v0/redis.py
@@ -86,7 +86,9 @@ class RedisRequires(Object):
             Dict: dict containing the app data.
         """
         relation = self.model.get_relation(self.relation_name)
-        return relation.data.get(relation.app)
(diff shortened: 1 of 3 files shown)


reneradoi:
Hello @alithethird , I tried to look into the issue with the failed integration test. It is failing in the build-step, probably because of the change to the lib. Unfortunately I cannot really debug that without cloning your private repo. Could you please create a branch on this repo here, and push your changes to this new branch? We can also have a session together if that would help you, just let me know.

alithethird:
Hi @reneradoi ! Thank you! I can not create a branch because I do not have write access. My fork should be private but just in case I added you as a collaborator to the fork.

reneradoi (changes requested):
Hey @alithethird, in general we'd be fine with adjusting this, as long as the interface stays the same. But after looking at the failing test, I don't think that this is flaky as you mentioned earlier. I have run the integration tests against the main branch on multiple tries, and there was no failure.

The error from the CI pipeline is [here](https://github.com/canonical/redis-k8s-operator/actions/runs/11340535060/job/31940363754?pr=100#step:6:146). I suggest you check this and adjust your code accordingly.

reneradoi:
@alithethird After debugging with a local deployment, I've found the cause of the failing tests and fixed it by updating the pylibjuju to a newer version. Tests are now stable again. Furthermore I've adjusted the logic for the lib changes a little to make them more robust. Please have another look if this is fine for you.

reneradoi (approved):

lib/charms/redis_k8s/v0/redis.py
```diff
@@ -99,9 +111,16 @@ def url(self) -> Optional[str]:
             str: the Redis URL.
         """
         relation_data = self.relation_data
+        app_data = self.app_data
         if not relation_data:
             return None
         redis_host = relation_data.get("hostname")
+
+        if app_data:
```
Mehdi-Bendriss:
```suggestion
        if not (relation_data := self.relation_data):
            return None
            
        redis_host = relation_data.get("hostname")

        if app_data := self.app_data:
```

Mehdi-Bendriss (approved):
Thanks! Left a minor comment

alithethird (commented):
Looks great! Thanks a lot for your time.

